PR #1, opened 2018-12-11
chore: Address some metadata

RomainMuller:
Correct some stuff in `README.md`, and add some missing metadata fields in `package.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ building, testing and publishing code libraries through AWS CodeBuild and AWS
 CodePipeline.
 
 This guide will walk you through setting up a delivlib pipeline and utilizing
-all it's features.
+all its features.
 
 A delivlib pipeline consists of the following sequential stages. Each stage will
 execute all tasks concurrently:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -436,10 +436,9 @@ to access old versions of the docs if they wish).
 
 NOTE: static website content can grow big. If you use the same repository for
 your source code and GitHub pages, your repo clones will get exponentially large
-due to the fact the GitHub Pages do not allow force pushing commits into a
-gh-pages branch. This means that we are storing each version both historically
-and vertically under the `${version}/` directory. It is recommended to publish
-docs into a separate repository.
+due to the fact that we are storing each version both historically and
+vertically under the `${version}/` directory. It is recommended to publish docs
```
eladb: Should we also instruct people to allow force pushing and modify our scripts to support that? (separate PR)

RomainMuller: Maybe. Right now I was mostly concerned about there being apparently wrong information about `gh-pages` and I don't want us to be seen as propagating fake news, but yeah maybe we should change to a model that squashes down to 1 commit and force-pushes at a subsequent PR.

PR #5, opened 2018-12-12
chore: unit test for `pipelineName`

eladb:
Our first unit test, to make it a habit to write
one for every feature.

Misc:
- Some updates to README
- Add "npm test synth" which outputs the template for inspection
- Clean up dashboard code that's not really useful (we can introduce later perhaps)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6, opened 2018-12-13
Raise an alarm when any stages are in a Failed state.

sam-goodwin:
A function runs every minute and calls GetPipelineState for the provided pipeline's name, counts the number of failed stages and emits a JSON log { failedCount: <number> }. A metric filter is then configured to track this value as a CloudWatch metric, and a corresponding alarm is set to fire when the maximim value of a single 5-minute interval is >= 1.

### DIFF
--- a/lib/pipeline-watcher.ts
+++ b/lib/pipeline-watcher.ts
@@ -1,0 +1,29 @@
+import AWS = require('aws-sdk');
+
+const pipeline = new AWS.CodePipeline();
+
+/**
+ * Lambda function for checking the stages of a CodePipeline and emitting log
+ * entries with { failedCount = <no. of failed stages> } for async metric
+ * aggregation via metric filters.
+ *
+ * It requires the pipeline's name be set as the 'pipelineName' environment variable.
+ */
+export async function handler() {
+  const pipelineName = process.env.pipelineName;
+  if (!pipelineName) {
+    throw new Error("Pipeline name expects environment variable: 'pipelineName'");
+  }
+
+  const state = await pipeline.getPipelineState({
+    name: pipelineName
+  }).promise();
+
+  if (state.stageStates) {
+    const failedCount = state.stageStates
+      .filter(stage => stage.latestExecution !== undefined && stage.latestExecution.status === 'Failed')
+      .length;
+
+    process.stdout.write(JSON.stringify({failedCount}));
+  }
+}
--- a/lib/pipeline-watcher/index.ts
+++ b/lib/pipeline-watcher/index.ts
@@ -1,0 +1,2 @@
+export * from './watcher-handler';
+export * from './watcher';
--- a/lib/pipeline-watcher/index.ts
+++ b/lib/pipeline-watcher/index.ts
@@ -1,2 +1,1 @@
-export * from './watcher-handler';
 export * from './watcher';
--- a/lib/pipeline-watcher/watcher-function.ts
+++ b/lib/pipeline-watcher/watcher-function.ts
@@ -7,23 +7,23 @@ const pipeline = new AWS.CodePipeline();
  * entries with { failedCount = <no. of failed stages> } for async metric
  * aggregation via metric filters.
  *
- * It requires the pipeline's name be set as the 'pipelineName' environment variable.
+ * It requires the pipeline's name be set as the 'PIPELINE_NAME' environment variable.
  */
 export async function handler() {
-  const pipelineName = process.env.pipelineName;
+  const pipelineName = process.env.PIPELINE_NAME;
   if (!pipelineName) {
-    throw new Error("Pipeline name expects environment variable: 'pipelineName'");
+    throw new Error("Pipeline name expects environment variable: 'PIPELINE_NAME'");
   }
 
   const state = await pipeline.getPipelineState({
     name: pipelineName
   }).promise();
 
+  let failedCount = 0;
   if (state.stageStates) {
-    const failedCount = state.stageStates
+    failedCount = state.stageStates
       .filter(stage => stage.latestExecution !== undefined && stage.latestExecution.status === 'Failed')
       .length;
-
-    process.stdout.write(JSON.stringify({failedCount}));
   }
+  process.stdout.write(JSON.stringify({failedCount}));
 }
--- a/lib/pipeline-watcher/watcher.ts
+++ b/lib/pipeline-watcher/watcher.ts
@@ -82,7 +82,7 @@ export class PipelineWatcher extends cdk.Construct {
     });
 
     this.alarm = new cloudwatch.Alarm(this, 'Alarm', {
-      alarmDescription: `Pipeline ${props.title || props.pipeline.pipelineName} has failed failed stages`,
+      alarmDescription: `Pipeline ${props.title || props.pipeline.pipelineName} has failed stages`,
       metric: new cloudwatch.Metric({
         metricName,
         namespace: metricNamespace,
--- a/lib/pipeline-watcher/watcher.ts
+++ b/lib/pipeline-watcher/watcher.ts
@@ -1,0 +1,97 @@
+import cloudwatch = require('@aws-cdk/aws-cloudwatch');
+import cpipeline = require('@aws-cdk/aws-codepipeline');
+import events = require('@aws-cdk/aws-events');
+import iam = require('@aws-cdk/aws-iam');
+import lambda = require('@aws-cdk/aws-lambda');
+import logs = require('@aws-cdk/aws-logs');
+import cdk = require('@aws-cdk/cdk');
+import fs = require('fs');
+import path = require('path');
+
+export interface PipelineWatcherProps {
+  /**
+   * Code Pipeline to monitor for failed stages
+   */
+  pipeline: cpipeline.Pipeline;
+
+  /**
+   * Set the pipelineName of the alarm description.
+   *
+   * Description is set to 'Pipeline <title> has failed stages'
+   *
+   * @default pipeline's name
+   */
+  title?: string;
+}
+
+/**
+ * Construct which watches a Code Pipeline for failed stages and raises an alarm
+ * if there are any failed stages.
+ *
+ * A function runs every minute and calls GetPipelineState for the provided pipeline's
+ * name, counts the number of failed stages and emits a JSON log { failedCount: <number> }.
+ * A metric filter is then configured to track this value as a CloudWatch metric, and
+ * a corresponding alarm is set to fire when the maximim value of a single 5-minute interval
+ * is >= 1.
+ */
+export class PipelineWatcher extends cdk.Construct {
+  public readonly alarm: cloudwatch.Alarm;
+
+  constructor(parent: cdk.Construct, name: string, props: PipelineWatcherProps) {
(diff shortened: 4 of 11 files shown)


lib/pipeline-watcher.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Lambda function for checking the stages of a CodePipeline and emitting log
+ * entries with { failedCount = <no. of failed stages> } for async metric
+ * aggregation via metric filters.
+ *
+ * It requires the pipeline's name be set as the 'pipelineName' environment variable.
+ */
+export async function handler() {
+  const pipelineName = process.env.pipelineName;
```
eladb: Convention: all caps for environment variables

lib/pipeline-watcher.ts
```diff
@@ -0,0 +1,29 @@
...
+    name: pipelineName
+  }).promise();
+
+  if (state.stageStates) {
+    const failedCount = state.stageStates
+      .filter(stage => stage.latestExecution !== undefined && stage.latestExecution.status === 'Failed')
+      .length;
+
+    process.stdout.write(JSON.stringify({failedCount}));
+  }
```
eladb: Just to be safe, I'd print `{ failedCount: 0 }` if `state.stageStates` is undefined.

lib/pipeline.ts
```diff
@@ -255,25 +259,94 @@ export class Pipeline extends cdk.Construct {
...
-    const pipelineFailureTopic = new sns.Topic(this, 'PipelineFailureTopic');
+  private addFailureAlarm(title?: string): cloudwatch.Alarm {
+    return new PipelineWatcher(this, 'PipelineWatcher', {
+      pipeline: this.pipeline,
+      title
+    }).alarm;
+  }
+}
+
+interface PipelineWatcherProps {
```
eladb: Code organization: perhaps put all pipeline watcher stuff under a `lib/pipeline-watcher` directory?

lib/pipeline.ts
```diff
@@ -255,25 +259,94 @@ export class Pipeline extends cdk.Construct {
...
+        pipelineName: props.pipeline.pipelineName
+      }
+    });
+
+    // See https://github.com/awslabs/aws-cdk/issues/1340 for exposing grants on the pipeline.
+    pipelineWatcher.addToRolePolicy(new iam.PolicyStatement()
+      .addResource(props.pipeline.pipelineArn)
+      .addAction('codepipeline:GetPipelineState'));
+
+    // Explicitly and pre-emptively create the function's log group because we need it for the metric filter.
```
eladb: There's a race condition here between Lambda and CloudFormation. I remember @rix0rrr had to create a custom resource to implement something like that and I wonder if that's the case, if we should actually include this in the Lambda construct library.

lib/pipeline.ts
```diff
@@ -255,25 +259,94 @@ export class Pipeline extends cdk.Construct {
...
+    const logGroup = new logs.LogGroup(this, 'Logs', {
+      logGroupName: new cdk.FnConcat('/aws/lambda/', pipelineWatcher.functionName).toString(),
```
eladb: We prefer not to use `FnConcat` explicitly and let the stringified token magic do it's trick:

    `/aws/lambda/${pipelineWatcher.functionName}`

lib/pipeline.ts
```diff
@@ -255,25 +259,94 @@ export class Pipeline extends cdk.Construct {
...
+    });
+
+    new events.EventRule(this, 'Trigger', {
+      scheduleExpression: 'rate(1 minute)',
+      targets: [pipelineWatcher]
+    });
+
+    const metricName = 'FailedStages';
+    // TODO: This creates a long namespace, better alternatives?
+    const metricNamespace = new cdk.FnConcat('CodePipeline/', props.pipeline.pipelineName).toString();
```
eladb: avoid `FnConcat`

lib/pipeline.ts
sam-goodwin: I can address this by having the EventRule trigger depend on the LogGroup Resource. This ensures the log group is created before the first invocation of the function. Lambda lazily creates the log group on first invocation, so no more race condition. The only down-side I can see is that I have to drop down to L1 and rely on the findChild(‘Resource’) convention.

```typescript
const logGroupResource = logGroup.findChild('Resource') as cdk.Resource;
const triggerResource = trigger.findChild('Resource') as cdk.Resource;
triggerResource.addDependency(logGroupResource);
```

lib/pipeline.ts
sam-goodwin: Sweet!

lib/pipeline.ts
eladb: I guess... 

We should try to find a better way to allow defining dependencies between resources (https://github.com/awslabs/aws-cdk/issues/95)

We have an issue to try and make log groups accessible in the CDK (https://github.com/awslabs/aws-cdk/issues/667).

lib/pipeline-watcher/watcher-function.ts
```diff
@@ -0,0 +1,29 @@
+import AWS = require('aws-sdk');
```
eladb: Sorry for the nit picking, but I'd call this "watcher-handler"

lib/pipeline-watcher/watcher-function.ts
```diff
@@ -0,0 +1,29 @@
...
+  }).promise();
+
+  let failedCount = 0;
+  if (state.stageStates) {
+    failedCount = state.stageStates
+      .filter(stage => stage.latestExecution !== undefined && stage.latestExecution.status === 'Failed')
+      .length;
+  }
+  process.stdout.write(JSON.stringify({failedCount}));
+}
```
eladb: Can we write a simple unit test for this handler? (I know we don't have any UT for handlers in this library, yet, but let's raise the bar)

lib/pipeline.ts
```diff
@@ -6,6 +6,7 @@ import iam = require('@aws-cdk/aws-iam');
 import sns = require('@aws-cdk/aws-sns');
 import cdk = require('@aws-cdk/cdk');
 import path = require('path');
+import { PipelineWatcher } from './pipeline-watcher/watcher';
```
eladb: Create an `pipeline-watcher/index.ts` file and `export * from './watcher'`, so here it will be:

```
import { PipelineWatcher } from './pipeline-watcher'
```

lib/pipeline-watcher/watcher.ts
```diff
@@ -0,0 +1,97 @@
...
+    new logs.MetricFilter(this, 'MetricFilter', {
+      filterPattern: logs.FilterPattern.exists('$.failedCount'),
+      metricNamespace,
+      metricName,
+      metricValue: '$.failedCount',
+      logGroup
+    });
+
+    this.alarm = new cloudwatch.Alarm(this, 'Alarm', {
+      alarmDescription: `Pipeline ${props.title || props.pipeline.pipelineName} has failed failed stages`,
```
eladb: "failed failed"

lib/pipeline-watcher/watcher.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const logGroupResource = logGroup.findChild('Resource') as cdk.Resource;
+    const triggerResource = trigger.findChild('Resource') as cdk.Resource;
+    triggerResource.addDependency(logGroupResource);
+
+    const metricName = 'FailedStages';
+    // TODO: This creates a long namespace, better alternatives?
+    const metricNamespace =  `CodePipeline/${props.pipeline.pipelineName}`;
+
+    new logs.MetricFilter(this, 'MetricFilter', {
```
eladb: You can probably use [`logGroup.extractMetric`](https://awslabs.github.io/aws-cdk/refs/_aws-cdk_aws-logs.html?highlight=metricfilter#metric-filters) here.

lib/pipeline-watcher/watcher.ts
```diff
@@ -0,0 +1,97 @@
...
+      filterPattern: logs.FilterPattern.exists('$.failedCount'),
+      metricNamespace,
+      metricName,
+      metricValue: '$.failedCount',
+      logGroup
+    });
+
+    this.alarm = new cloudwatch.Alarm(this, 'Alarm', {
+      alarmDescription: `Pipeline ${props.title || props.pipeline.pipelineName} has failed failed stages`,
+      metric: new cloudwatch.Metric({
```
eladb: I am surprised we don't have `metricFilter.metric()`, can you raise an issue?

lib/pipeline-watcher/watcher-function.ts
sam-goodwin: Working on it .. it's a pain since we use nodeunit (deprecated) and most of what the handler does is create a client, check env variables, make a request and log to stdout - all of which are annoying to test, ha.

lib/pipeline.ts
sam-goodwin: Another alternative is that I have the handler call `putMetricData` directly instead of using a MetricFilter. Let me know what you think so I can wrap this up.

lib/pipeline-watcher/watcher-function.ts
eladb: See #5 - I don't think we need to use nodeunit here. Moving to jest?

lib/pipeline.ts
eladb: I think the metric filter route is the right one, just want to make sure we have a better path in the CDK for future generations to use it. Tactically, your solution for delivlib is perfectly fine.

lib/pipeline-watcher/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './watcher-handler';
```
eladb: I don't think we should export the handler, just the construct. If you want to import it in your unit test you can just use '../lib/pipeline-watcher/watcher-handler'.

lib/pipeline-watcher/watcher.ts
sam-goodwin: Can't use dimensions in metric filters unfortunately.

lib/pipeline-watcher/watcher.ts
```diff
@@ -0,0 +1,97 @@
...
+      targets: [pipelineWatcher]
+    });
+
+    const logGroupResource = logGroup.findChild('Resource') as cdk.Resource;
+    const triggerResource = trigger.findChild('Resource') as cdk.Resource;
+    triggerResource.addDependency(logGroupResource);
+
+    const metricName = 'FailedStages';
+    // TODO: This creates a long namespace, better alternatives?
+    const metricNamespace =  `CodePipeline/${props.pipeline.pipelineName}`;
```
sam-goodwin: @rix0rrr I've called it out in the code here. Any other suggestions for distinguishing individual pipeline's metrics from each other? I can't use dimensions in metric filters :(

lib/pipeline-watcher/watcher.ts
sam-goodwin: I could instead make the metric name ugly instead of the namespace:

* Namespace: `CDK/Delivlib`
* MetricName: `<pipeline-name>_FailedStages`

PR #9, opened 2018-12-13
feat(pipeline): concurrency limit

eladb:
If `concurrency` is set to a number, no more than N actions will be run concurrently in the TEST and PUBLISH stages.

This allows overcoming the default account limitation of a 1 concurrency.

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -82,6 +82,16 @@ export interface PipelineProps {
    * @default true
    */
   restartExecutionOnUpdate?: boolean;
+
+  /**
+   * Enables tests and publishers to be executed concurrently in the pipeline.
+   *
+   * You can set this to "false" if, for example, your account has an AWS CodeBuild
+   * concurrency limitation.
+   *
+   * @default true
+   */
+  concurrency?: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


RomainMuller:
I would rather:
1. Make concurrency a `number` that defaults to 10
2. Keep entities in the same stage, but specify `runOrder`every `concurrency`th action

This avoids you runnning *so fast* against the stage count limit and still controls concurrency of the CodeBuild. Ideally, concurrency would even be something that can be controlled on the stages itself.

eladb:
oh, right, totally forgot about `runOrder`. Let me look into that. I would like the default to be fully concurrent though.

lib/pipeline.ts
```diff
@@ -82,6 +83,16 @@ export interface PipelineProps {
...
    */
   restartExecutionOnUpdate?: boolean;
+
+  /**
+   * Indicates the concurrency limit test and publish stages.
+   *
+   * For example, if this value is 2, then only two actions will execute concurrently.
+   * If this value is 1, the pipeline will not have any concurrent execution.
+   *
+   * @default no limit
```
RomainMuller: The reason I had mentioned a default of 10 is because it used to be the CodePipeline limit for parallel actions in a stage. But in fact, this was increased to 50 recently. I would tend to default to 50 for that reason, but I'm also happy with "no limit".

PR #14, opened 2018-12-17
fix: Logger reference in CSC custom resources

RomainMuller:
The lambda code that back the Code Signing Certificate resources were
referencing the `logging` module using an incorrect name. This fixes the
incorrect naming, and adds descriptions to the Lambda functions (as well
as tagging a `purpose` on the entities that can be safely replaced).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/code-signing/certificate-signing-request/index.py
+++ b/lib/code-signing/certificate-signing-request/index.py
@@ -23,116 +23,116 @@ CFN_SUCCESS = "SUCCESS"
 CFN_FAILED = "FAILED"
 
 def handle_event(event, aws_request_id):
-  import boto3, subprocess, tempfile
+   import boto3, subprocess, tempfile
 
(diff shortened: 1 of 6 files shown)


lib/code-signing/certificate-signing-request/index.py
```diff
@@ -107,33 +107,32 @@ def main(event, context):
 #---------------------------------------------------------------------------------------------------
 # sends a response to cloudformation
 def cfn_send(event, context, responseStatus, responseData={}, physicalResourceId=None, noEcho=False, reason=None):
+  responseUrl = event['ResponseURL']
```
eladb: I believe pythonians like 3 spaces...

RomainMuller: Hashtag weird people 😁

PR #17, opened 2018-12-17
feat(pgp-secret): Surface parameterName attribute

RomainMuller:

### DIFF
--- a/lib/pgp-secret.ts
+++ b/lib/pgp-secret.ts
@@ -61,6 +61,7 @@ interface PGPSecretProps {
  */
 export class PGPSecret extends cdk.Construct {
   public readonly secretArn: string;
+  public readonly parameterName: string;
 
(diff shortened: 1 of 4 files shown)


lib/pgp-secret.ts
```diff
@@ -61,6 +61,7 @@ interface PGPSecretProps {
  */
 export class PGPSecret extends cdk.Construct {
   public readonly secretArn: string;
+  public readonly parameterName: string;
```
eladb: Shouldn't we call this `publicKeyParameterName`?

eladb:
Can you add a unit test please?

lib/pgp-secret.ts
RomainMuller: No - in fact I'll even be adding a shared interface for `PGPSecret` and `PrivateKey` (from `code-signing`) to make usage of those types of secrets more consistent (that's particularly handy when you need to share those across stacks in  a streamlined manner).

PR #20, opened 2018-12-18
feat: Allow update of PGPSecret and PrivateKey

RomainMuller:
Updated the python code that backs the custom resource implementations for the
PGPSecret and PrivateKey constructs so they allow updating the KMS encryption
keys used by AWS SecretsManager without necessarily re-generating new secrets.

Also, added a `resourceVersion` has to the properties so the resource handler is
called again if the implementation code changed.

BREAKING CHANGE: this also changes the API of the PGPSecret and
CodeSigningCertificate constructs to offer a consistent API for accessing the name
and ARNs of the secret and parameters associated with the secrets, through the
`ICredentialPair` interface.


lib/code-signing/code-signing-certificate.ts
```diff
@@ -97,23 +81,22 @@ export class CodeSigningCertificate extends cdk.Construct implements ICodeSignin
...
    */
-
-  public readonly privateKeySecretVersionId: string;
+  public readonly secretVersionId: string;
 
   /**
-   * The name of the AWS SSM parameter that holds the certificate for this CSC.
+   * The ARN of the AWS SSM Parameter that holds the certificate for this CSC.
    */
-  public readonly certificateParameterName: string;
```
eladb: I thought the previous names were clearer. `parameter` and `secret` are the mechanisms in which you store these, not the actual intent

lib/code-signing/private-key.ts
```diff
@@ -63,28 +63,27 @@ export class RsaPrivateKeySecret extends cdk.Construct {
...
+      code: new lambda.AssetCode(codeLocation),
       timeout: 300,
     });
 
     this.secretArnLike = cdk.ArnUtils.fromComponents({
       service: 'secretsmanager',
       resource: 'secret',
       sep: ':',
-      resourceName: `${props.secretName}-*`
+      resourceName: `${props.secretName}-??????`
```
eladb: What's this about? Does it actually work?

lib/credential-pair.ts
```diff
@@ -0,0 +1,35 @@
...
+ * - A username and a password
+ * - A private key and a certificate
+ * - An OpenPGP Private key and its public part
+ */
+export interface ICredentialPair {
+  /**
+   * The ARN of the SSM parameter containing the public part of this credential
+   * pair.
+   */
+  readonly parameterArn: string;
```
eladb: Again, the word "public" and "private" should appear in these property names. Ideally the type should not be just strings but rather `ssm.IParameter` and `secretsmanager.ISecret`, but I don't think we have those, right?

lib/pgp-secret.ts
```diff
@@ -91,17 +104,21 @@ export class PGPSecret extends cdk.Construct {
     const secret = new cfn.CustomResource(this, 'Resource', {
       lambdaProvider: fn,
       properties: {
+        resourceVersion: hashFileOrDirectory(codeLocation),
```
eladb: raise an issue to add this as a feature to CustomResource

lib/util.ts
```diff
@@ -11,3 +15,25 @@ export function determineRunOrder(index: number, concurrency?: number) {
...
+
+/**
+ * Hashes the contents of a file or directory. If the argument is a directory,
+ * it is assumed not to contain symlinks that would result in a cyclic tree.
+ *
+ * @param fileOrDir the path to the file or directory that should be hashed.
+ *
+ * @returns a SHA256 hash, base-64 encoded.
+ */
+export function hashFileOrDirectory(fileOrDir: string): string {
```
eladb: Raise an issue to add this to `FileAsset`

lib/code-signing/code-signing-certificate.ts
RomainMuller: Yeah but I still want to have a common interface for the PGPSecret and PrivateKey...

lib/code-signing/private-key.ts
RomainMuller: In fact this is what the AWS documentation for SecretsManager recommends. The `?` in ARN patterns matches exactly one character, and secret arms are composed with the name to which `-` and 6 random characters are added. That is the "safe" way to grant permissions to a secret before it is created (or if you can't pass around the actual full ARN). It has been tested to work.

lib/code-signing/code-signing-certificate.ts
eladb: Both have a "private part" and a "public part", no?

lib/credential-pair.ts
RomainMuller: Yeah I don't reckon we have those either.

lib/util.ts
RomainMuller: awslabs/aws-cdk#1400

lib/pgp-secret.ts
RomainMuller: awslabs/aws-cdk#1401

custom-resource-handlers/src/certificate-signing-request.ts
```diff
@@ -0,0 +1,113 @@
...
+  case cfn.RequestType.UPDATE:
+    return _createSelfSignedCertificate(event);
+  case cfn.RequestType.DELETE:
+    // Nothing to do - this is not a "Physical" resource
+    return { CSR: '', SelfSignedCertificate: '' };
+  }
+}
+
+async function _createSelfSignedCertificate(event: cfn.Event): Promise<ResourceAttributes> {
+  const tempDir = await util.promisify(fs.mkdtemp)(path.join(os.tmpdir(), 'x509CSR-'));
```
eladb: A more common pattern is to invoke `promisify` immediately after the `import` statements and reuse the result:

```ts
import { promisify } from 'util';
import fs = require('fs')l;

const mkdtemp = promisify(fs.mkdtemp);
const readFile = promisify(fs.readFile);
// ...
```

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+const ssm = new aws.SSM();
+
+export async function main(event: cfn.Event, context: lambda.Context): Promise<void> {
+  try {
+    // tslint:disable-next-line:no-console
+    console.log(`Input event: ${JSON.stringify(event)}`);
+    const attributes = await handleEvent(event, context);
+    await cfn.sendResponse(event,
+                           cfn.Status.SUCCESS,
+                           attributes.SecretArn || event.PhysicalResourceId || context.logStreamName,
```
eladb: Feels like if `SecretArn` is undefined/empty/null, it's a bug, isn't it?

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+                           e.message);
+  }
+}
+
+interface ResourceAttributes {
+  SecretArn: string;
+  SecretVersionId?: string;
+  ParameterName?: string;
+
+  [name: string]: string | undefined;
```
eladb: Why open up this interface? Don't we have a well-known schema here?

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+        console.log(`New key required: ${key} changed from ${oldProps[key]} to ${props[key]}`);
+        newKey = true;
+      }
+    }
+  }
+
+  switch (event.RequestType) {
+  case cfn.RequestType.CREATE:
+  case cfn.RequestType.UPDATE:
+    return newKey
```
eladb: Just to make sure I understand, if we are in `newKey` mode (and UPDATE), the physical resource ID (secret ARN) will change and then CFN will send a DELETE on the old secret will be deleted after the new one has been created. correct? Maybe worth a short comment.

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+          ? await _createNewKey(event, context)
+          : await _updateExistingKey(event as cfn.UpdateEvent, context);
+  case cfn.RequestType.DELETE:
+    return await _deleteSecret(event);
+  }
+}
+
+async function _createNewKey(event: cfn.CreateEvent | cfn.UpdateEvent, context: lambda.Context): Promise<ResourceAttributes> {
+  const passPhrase = crypto.randomBytes(32).toString('base64');
+  const tempDir = await util.promisify(fs.mkdtemp)(path.join(os.tmpdir(), 'OpenPGP-'));
```
eladb: hoist all promisify() calls.

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    process.env.GNUPGHOME = tempDir;
+
+    const keyConfig = path.join(tempDir, 'key.config');
+    await util.promisify(fs.writeFile)(keyConfig, [
+      'Key-Type: RSA',
+      `Key-Length: ${event.ResourceProperties.KeySizeBits}`,
```
eladb: Add some validation that all required properties are passed in.

custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -0,0 +1,135 @@
...
+      KmsKeyId: event.ResourceProperties.KeyArn,
+      SecretString: keyMaterial,
+    };
+    const secret = event.RequestType === cfn.RequestType.CREATE
+                 ? await secretsManager.createSecret({ ...secretOpts, Name: event.ResourceProperties.SecretName }).promise()
+                 : await secretsManager.updateSecret({ ...secretOpts, SecretId: event.PhysicalResourceId }).promise();
+    await ssm.putParameter({
+      Description: `Public part of OpenPGP key ${secret.ARN} (version ${secret.VersionId})`,
+      Name: event.ResourceProperties.ParameterName,
+      Overwrite: event.RequestType === 'Update',
```
eladb: use `cfn.RequestType`

custom-resource-handlers/src/_rmrf.ts
```diff
@@ -0,0 +1,15 @@
+import fs = require('fs');
+import path = require('path');
+import util = require('util');
+
+export = async function _rmrf(filePath: string): Promise<void> {
+  const stat = await util.promisify(fs.stat)(filePath);
```
eladb: Hoist promisify() near imports

build-custom-resource-handlers.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
eladb: I wonder if a better way to organize this code is to use a lerna project with a private module that contains the runtime code (see [cdk-tweet-queue](https://github.com/eladb/cdk-tweet-queue) as an example). It will make it much easier to control reuse, use dependencies (fs-extra is called for), test in isolation, distinguish internal/external etc.

At the same time, you could then put all the general purpose Lambda-CFN stuff into yet-another module in the lerna repo and publish it separately for the world to consume and reuse, or consume something like [cfn-lambda](https://www.npmjs.com/package/cfn-lambda) - why reinvent the wheel?

test/custom-resource-handlers/_rmrf.test.ts
```diff
@@ -0,0 +1,11 @@
+import fs = require('fs');
+import os = require('os');
+import path = require('path');
+
+import _rmrf = require('../../custom-resource-handlers/src/_rmrf');
+
+test('resmoves a full directory', () => {
```
eladb: resmoves

custom-resource-handlers/src/_rmrf.ts
```diff
@@ -0,0 +1,15 @@
+import fs = require('fs');
+import path = require('path');
+import util = require('util');
+
+export = async function _rmrf(filePath: string): Promise<void> {
```
eladb: fs-extra?

test/custom-resource-handlers/private-key.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  await expect(mockSecretsManager.deleteSecret).not.toBeCalled();
+  return expect(cfn.sendResponse)
+    .toBeCalledWith(event,
+                    cfn.Status.FAILED,
+                    secretArn,
+                    { SecretArn: '' },
+                    expect.stringContaining('The KeySize property cannot be updated'));
+});
+
+test('Update (changing KeySize)', async () => {
```
eladb: Looks like a dup test

build-custom-resource-handlers.sh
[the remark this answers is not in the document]
RomainMuller: Can you configure a different output directory for a set of files?

custom-resource-handlers/src/_exec.ts
```diff
@@ -0,0 +1,19 @@
+import childProcess = require('child_process');
+import process = require('process');
+
+export = function _exec(command: string, ...args: string[]): Promise<string> {
```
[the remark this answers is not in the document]
RomainMuller: The module is the function. Given the file name and all, I figured it made more sense.

custom-resource-handlers/src/_exec.ts
```diff
@@ -0,0 +1,19 @@
...
+  return new Promise<string>((ok, ko) => {
+    const child = childProcess.spawn(command, args, { env: process.env, shell: false, stdio: ['ignore', 'pipe', 'inherit'] });
```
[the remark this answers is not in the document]
RomainMuller: When I didn't pass this, then GPGHOME didn't appear to be set... 🤷🏻‍♂️

custom-resource-handlers/src/_lambda.ts
```diff
@@ -0,0 +1,90 @@
+/**
```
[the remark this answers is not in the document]
RomainMuller: Then I have to make sure I got the types absolutely right (the documentation isn't great about what is always present, and what isn't).

custom-resource-handlers/src/certificate-signing-request.ts
```diff
@@ -0,0 +1,113 @@
...
+const secretsManager = new aws.SecretsManager();
+
+export async function main(event: cfn.Event, context: lambda.Context): Promise<void> {
+  try {
+    // tslint:disable-next-line:no-console
+    console.log(`Input event: ${JSON.stringify(event)}`);
+    const attributes = await handleEvent(event, context);
+    await cfn.sendResponse(event,
+                          cfn.Status.SUCCESS,
+                          event.LogicalResourceId || event.PhysicalResourceId || context.logStreamName,
```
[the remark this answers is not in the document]
RomainMuller: Means I have to pass it the context, which is used *only* for that... Sounds heavy-handed...

custom-resource-handlers/src/certificate-signing-request.ts
```diff
@@ -0,0 +1,113 @@
...
+                          attributes);
+  } catch (e) {
+    // tslint:disable-next-line:no-console
+    console.error(e);
+    await cfn.sendResponse(event,
+                          cfn.Status.FAILED,
+                          event.LogicalResourceId,
+                          { CSR: '', SelfSignedCertificate: '' },
```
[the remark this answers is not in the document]
RomainMuller: CloudFormation allowed me to UPDATE a DELETED resource, so it wouldn't surprise me if it allowed me to GetAtt on a failed resource somehow.

build-custom-resource-handlers.sh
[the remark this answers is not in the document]
eladb: As mentioned somewhere else, I think we should just move these into a separate module and configure it like any other module

custom-resource-handlers/src/pgp-secret.ts
RomainMuller: Successful DELETE has a blank SecretArn.

custom-resource-handlers/src/pgp-secret.ts
RomainMuller: TypeScript won't let this pass into `sendResponse` without this, because it does not conform to `{ [name: string]: string | undefined }`, for some reason, and I don't want to have the data argument be `any`.

custom-resource-handlers/src/pgp-secret.ts
RomainMuller: This felt obvious enough to me to not warrant a comment. But okay

build-custom-resource-handlers.sh
RomainMuller: Because this change already blew *way* beyond what was strictly necessary.

test/custom-resource-handlers/private-key.test.ts
RomainMuller: That is what you get for having your lunch break right after copying a test with intention of modifying it for a variant (but before having modified it).

custom-resource-handlers/src/_rmrf.ts
RomainMuller: If I take dependencies, I have to be certain they don't (and will never) include native code (otherwise I need to make sure it'd built for the Lambda container OS/architecture).

eladb: Makes sense

custom-resource-handlers/src/pgp-secret.ts
RomainMuller: Fun stuff - there actually is an issue here in that the old & new secret both have the same name, so the CLEANUP will cause the secret to be deleted. Not nice.

PR #25, opened 2018-12-19
fix: do not assume executable permissions on publishing scripts

eladb:
since codebuild does not preserve permissions, we can't assume
our publishing scripts are executable. this change either
chmods them at runtime or invokes them with `/bin/bash`

*Issue #, if available:*

*Description of changes:*


lib/publishing/maven/publish.sh
```diff
@@ -15,4 +15,6 @@ credentials=$(aws secretsmanager get-secret-value --secret-id ${MAVEN_LOGIN_SECR
 export MAVEN_USERNAME=$(node -e "console.log(${credentials}.username);")
 export MAVEN_PASSWORD=$(node -e "console.log(${credentials}.password);")
 
+chmod +x $SCRIPT_DIR/with-signing-key.sh
```
RomainMuller: Why do we `chmod` those, and in other places we just shell out to `bash`?

eladb: It's actually consistent with [the other usage](https://github.com/awslabs/aws-delivlib/blob/972a1c9ea61b2464f36204a32bc264f0641463ee/lib/publishing/github/publish.sh#L42) we have for `with-signing-key`.

PR #38, opened 2019-01-11
feat: wrap the superchain inmage in a Superchain construct.

### LINKED ISSUE #36: Cleanly expose `superchain` image
It's currently ugly if you want to use the `superchain` docker image:

```ts
new delivlib.LinuxPlatform(cbuild.LinuxBuildImage.fromAsset(this, 'superchain', {
  directory: path.join(__dirname, '..', 'node_modules', 'aws-delivlib', 'superchain')
}));
```

Should probably be something like:

```ts
new delivlib.SuperChain(this, 'superchain');
```

sam-goodwin:
Fixes #36 

Wrap the superchain inmage in a Superchain construct to simplify consumption


lib/superchain.ts
```diff
@@ -0,0 +1,19 @@
...
+import path = require('path');
+import { LinuxPlatform } from './shellable';
+
+/**
+ * Expose the superchain docker linux build image as a construct.
+ */
+export class Superchain extends cdk.Construct {
+  public readonly platform: LinuxPlatform;
+
+  constructor(scope: cdk.Construct, id: string) {
```
eladb: I am wondering why this actually needs an `id`. It should probably be a stack-singleton. I don't see any reason in the world for having more than one instance of those in the stack, so maybe we can implement it like this:

```ts
class Superchain extends LinuxPlatform {
  constructor(scope: cdk.Construct) {
    const stack = cdk.Stack.find(scope);
    const id = '<uuid>';
    const asset = stack.findChild(id) || cbuild.LinuxBuildImage.fromAsset(stack, id, { directory... });
    super(asset);
  }
}
```

Then, it can just be used as a LinuxPlatform.

sam-goodwin: Cool! I was wondering what the best approach to this would be - completely agree.

PR #39, opened 2019-01-11
Append / to tmpdir in test - was failing on linux

sam-goodwin:
Fixed a test that was failing on linux but not mac (for some reason).

Tested on both platforms and it runs successfully.


sam-goodwin:
Discarding this, was fixed as part of another PR.

PR #40, opened 2019-01-11
add a Canary construct

### LINKED ISSUE #37: Expose a `Canary` construct
Expose a construct, similar to `Shellable` or `Testable` that runs a script regularly and alarms on failures. 

Related to retries: https://github.com/awslabs/aws-delivlib/issues/34

sam-goodwin:
TODO: Fix the integ test - stack deployment is broken.

Fixes #37 

Adds a canary construct which regularly runs a script as a CodeBuild project and alarms on failures.


lib/canary.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(scope: cdk.Construct, id: string, props: CanaryProps) {
+    super(scope, id);
+
+    const shellable = new Shellable(this, 'Shellable', {
+      platform: props.platform || ShellPlatform.LinuxUbuntu,
+      scriptDirectory: props.scriptDirectory,
+      entrypoint: props.entrypoint,
+      environmentVariables: props.environmentVariables,
+      source: new cbuild.NoSource()
```
rix0rrr: Is this not the default for `Shellable`?

Can we not pass in the whole of `props` into `Shellable` like lazy people?

lib/canary.ts
```diff
@@ -0,0 +1,93 @@
...
+ * for failures. Ideal for running 'canary' scripts.
+ */
+export class Canary extends cdk.Construct {
+  public readonly alarm: cloudwatch.Alarm;
+
+  constructor(scope: cdk.Construct, id: string, props: CanaryProps) {
+    super(scope, id);
+
+    const shellable = new Shellable(this, 'Shellable', {
+      platform: props.platform || ShellPlatform.LinuxUbuntu,
```
rix0rrr: Is this not the default for `Shellable`?

lib/canary.ts
sam-goodwin: Looks like we can. Done.

test/expected.json
```diff
@@ -3,7 +3,7 @@ Resources:
     Type: AWS::S3::Bucket
     DeletionPolicy: Retain
     Metadata:
-      aws:cdk:path: delivlib-test/CodeCommitPipeline/BuildPipeline/ArtifactsBucket/Resource
```
eladb: revert ?

sam-goodwin: We can revert after 7 days - we have to wait for the secrets to expire - secret manager has a terrible development experience because you can't delete secrets immediately. I chose to check in an `expected.json` file that is compatible with `npm test update` since I ran into a bunch of problems. But maybe that was just because I was confused (user error) by the duplicate pipelines and won't happen again now that I've experienced the pain. Happy with whichever, let me know.

PR #45, opened 2019-01-17
fix: pgp-secret did not store passphrase in secrets manager

eladb:
keys are created with a passphrase, and "with-signing-key.sh"
scripts expect a "Passphrase" field in the secret which should be
encoded as JSON. 

Not sure how this actually ever worked.


RomainMuller:
This probably worked only because all of my tests were happening on a secret that already existed & then basically carefully verified the existing secret didn't get replaced... I guess that's why I'd prefer a test stack we can spin up & tear down each time, instead of keeping it up.

PR #46, opened 2019-01-17
refactor: improvements to shellable, testable and canary

eladb:
eventually testable and canary both utilize shellable, so modified
their props interface to extend `ShellableProps`. also, extracted
all optional props from `ShellableProps` to `ShellableOptions`, to
allow leaky APIs such as `addTest` to expose options to users.

modified the type of `environmentVariables` to key/value strings.
using ssm parameters doesn't seem like a very useful feature, especially
when it's possible to do that through CloudFormation.

add `pipline.addCanary` convenience method.

remove Testable. It's basically the same as Shellable with a default
script name. Redundant abstraction.

set IS_CANARY by default in canaries.

BREAKING CHANGE: `Testable` has been removed, `environmentVariables`
has been renamed to `env` and changed schema.

*Issue #, if available:*

*Description of changes:*


CONTRIBUTING.md
```diff
@@ -72,7 +72,17 @@ You can use the following npm scripts to manage the pipeline:
...
-To release a new delivlib version, run:
+Before releasing a new version, you first run an integration test to ensure that the various features are working.
+
+Here is how to execute an integration test:
+
+1. Setup credentials to our test AWS account: 712950704752
+2. Execute `npm test update`. This will update the integration test infrastructure. Wait until deployment has finished.
+3. Now, you will need to exercise the pipeline:
+   1. Clone https://github.com/awslabs/aws-delivlib-sample to your machine
+   2.
```
RomainMuller:
```
2. ...
3. Profit!
```

test/run-test.sh
```diff
@@ -22,7 +19,15 @@ fi
 npx cdk --no-version-reporting -a ${cdk_app} synth > ${actual}
 
 if [ "${1:-}" == "update" ]; then
+  hash="$(cat ${actual} | md5sum | cut -c1-6 | xargs)"
+  export TEST_STACK_NAME="delivlib-test-${hash}"
   npx cdk --no-version-reporting -a ${cdk_app} deploy ${2:-} ${3:-} ${4:-}
+  echo "Stack deployed, now, go to the console and wait for the pipeline to fully stabalize"
+  echo "Press ENTER to confirm that pipeline is green"
+  read
+  echo "Okay, now go to CFN console and delete the test stack ${TEST_STACK_NAME}"
```
RomainMuller: Would be sweet if we could actually... you know... go and check that the pipeline is, indeed, green. We can have this a TODO / separate issue, though.

test/run-test.sh
```diff
@@ -22,7 +19,15 @@ fi
 npx cdk --no-version-reporting -a ${cdk_app} synth > ${actual}
 
 if [ "${1:-}" == "update" ]; then
+  hash="$(cat ${actual} | md5sum | cut -c1-6 | xargs)"
+  export TEST_STACK_NAME="delivlib-test-${hash}"
   npx cdk --no-version-reporting -a ${cdk_app} deploy ${2:-} ${3:-} ${4:-}
+  echo "Stack deployed, now, go to the console and wait for the pipeline to fully stabalize"
+  echo "Press ENTER to confirm that pipeline is green"
```
RomainMuller: You're not telling me what I should do if it isn't 😱

lib/canary.ts
```diff
@@ -3,72 +3,38 @@ import cbuild = require('@aws-cdk/aws-codebuild');
 import events = require('@aws-cdk/aws-events');
 import cdk = require('@aws-cdk/cdk');
 
-import { Shellable, ShellPlatform } from './shellable';
+import { Shellable, ShellableProps } from './shellable';
 
-export interface CanaryProps {
+export interface CanaryProps extends ShellableProps {
```
sam-goodwin: The reason I didn't extend `ShellableProps` is because a canary always has no source, where `Shellable` lets you configure it.

lib/pipeline.ts
```diff
@@ -154,15 +155,24 @@ export class Pipeline extends cdk.Construct {
...
-    const test = new Testable(this, id, props);
-    test.addToPipeline(stage, this.buildOutput, this.determineRunOrderForNewAction(stage));
+    const test = new Shellable(this, id, props);
+    test.addToPipeline(stage, `Test${id}`, this.buildOutput, this.determineRunOrderForNewAction(stage));
 
     this.addBuildFailureNotification(test.project, `Test ${id} failed`);
   }
 
+  /**
+   * Convinience/discovery method that defines a canary test in your account.
```
sam-goodwin: 'Convenience'

lib/pipeline.ts
```diff
@@ -154,15 +155,24 @@ export class Pipeline extends cdk.Construct {
...
+   * @param id the construct id
+   * @param props canary options
+   */
+  public addCanary(id: string, props: CanaryProps) {
+    return new Canary(this, `Canary${id}`, props);
```
sam-goodwin: Could group canaries under a canary scope: `new cdk.Construct(this, 'Canary')`;

lib/publishing.ts
```diff
@@ -48,11 +48,11 @@ export class PublishToMavenProject extends cdk.Construct implements IPublisher {
       platform: new LinuxPlatform(cbuild.LinuxBuildImage.UBUNTU_14_04_NODEJS_10_1_0),
       scriptDirectory: path.join(__dirname, 'publishing', 'maven'),
       entrypoint: 'publish.sh',
-      environmentVariables: {
-        STAGING_PROFILE_ID: { value: props.stagingProfileId },
-        SIGNING_KEY_SCOPE: { value: props.signingKey.scope },
-        FOR_REAL: { value: forReal },
-        MAVEN_LOGIN_SECRET: { value: props.mavenLoginSecret.secretArn }
+      env: {
```
sam-goodwin: `env` doesn't seem to be consistent with the naming in other constructs. Maybe `environment`?

lib/publishing.ts
```diff
@@ -48,11 +48,11 @@ export class PublishToMavenProject extends cdk.Construct implements IPublisher {
...
+        STAGING_PROFILE_ID: props.stagingProfileId,
+        SIGNING_KEY_SCOPE: props.signingKey.scope,
+        FOR_REAL: forReal,
```
sam-goodwin: For real dawg

lib/pipeline.ts
RomainMuller: Reads the same to a frenchman. 💁🏻‍♂️

test/run-test.sh
eladb: Totally could. Totally in a separate issue.

test/run-test.sh
eladb: "Go figure" would probably be the recommendation.

test/run-test.sh
eladb: #49

PR #47, opened 2019-01-17
[WIP] feat(shallable): assume-role

eladb:
Allow users to assume a different role in shellable actions.

Remaining work:

- [ ] Unit test
- [ ] Integration test


lib/shellable.ts
```diff
@@ -50,6 +50,74 @@ export interface ShellableOptions {
...
+   * logged by the account that owns the role.  The role session name is also
+   * used in the ARN of the assumed role principal. This means that subsequent
+   * cross-account API requests using the tem- porary security credentials will
+   * expose the role session name to the external account in their CloudTrail
+   * logs.
+   *
+   * The regex used to validate this parameter is a string of characters
+   * consisting  of upper- and lower-case alphanumeric characters with no
+   * spaces. You can also include underscores or any of the following
+   * characters: =,.@-
```
sam-goodwin: Provide the regex?

lib/shellable.ts
```diff
@@ -117,14 +185,20 @@ export class Shellable extends cdk.Construct {
...
       }
     });
 
     this.role = this.project.role;
     asset.grantRead(this.role);
+
+    if (props.assumeRole && this.project.role) {
+      this.project.role.addToPolicy(new iam.PolicyStatement()
+        .addAction('sts:AssumeRole')
+        .addResource(props.assumeRole.roleArn));
```
sam-goodwin: Should you include the external id condition if it exists?

lib/shellable.ts
eladb: Copied this from AWS docs... Rephrasing. Why would the user care that there's a regex validating their input?

lib/shellable.ts
eladb: The condition should be added by the destination account, not the assuming account. I don't think this would have any effect from this side.

PR #65, opened 2019-02-11
fix: Add missing permission to PGPSecret CustomResource

RomainMuller:
On `UPDATE`, the CustomResource needs to be able to re-read the private
key, so it is able to produce the public part of it again. This adds the
missing permission.


lib/pgp-secret.ts
```diff
@@ -87,8 +87,9 @@ export class PGPSecret extends cdk.Construct implements ICredentialPair {
       initialPolicy: [
         new iam.PolicyStatement()
           .addActions('secretsmanager:CreateSecret',
-                      'secretsmanager:UpdateSecret',
                       'secretsmanager:DeleteSecret',
+                      'secretsmanager:GetSecretValue',
```
eladb: unit test?

PR #67, opened 2019-02-12
feat: Rename PGPSecret to OpenPGPKeyPair

RomainMuller:
- Deprecates the use of the `PGPKey` class in favor of using the
  `OpenPGPKeyPair` class directly.
- Updated the publishers that consume OpenPGP keys to use
  `OpenPGPKeyPair`s.
- Made the `encryptionKey` property of `OpenPGPKeyPair` optional,
  allowing use of the "default" key.

BREAKING CHANGE: The `PGPSecret` class was renamed to `OpenPGPKeyPair`.

Prepares: awslabs/cdk-ops#240


lib/open-pgp-key-pair.ts
```diff
@@ -84,22 +83,34 @@ export class PGPSecret extends cdk.Construct implements ICredentialPair {
...
+                  'secretsmanager:DeleteSecret',
+                  'secretsmanager:GetSecretValue',
+                  'secretsmanager:UpdateSecret')
+      .addResource(cdk.Stack.find(this).formatArn({
+        service: 'secretsmanager',
+        resource: 'secret',
+        sep: ':',
+        resourceName: `${props.secretName}-??????`
+      })));
+    fn.addToRolePolicy(new iam.PolicyStatement()
```
eladb: please add some whitespace between the statements

RomainMuller: Done! Somehow I thought I had those whitespaces... Must have been lost in some of my touchup iterations :)

PR #70, opened 2019-02-14
fix: Don't attempt deleting OpenPGP keys' secrets

RomainMuller:
Since the grants on the SSM and SecretsManager API are tightened down to just the particular secrets/parameters being managed (given their names), the permission to "delete" the secret no longer exists when the resource is called if it no longer is in the stack (for the permission grant also is no longer in the stack).


RomainMuller:
@rix0rrr the removal of `await secretsManager.deleteSecret({ SecretId: event.PhysicalResourceId }).promise();` results in something no longer being deleted...

rix0rrr:
Well in the revision that I saw it was being reformatted, not removed

PR #72, opened 2019-02-18
fix: Upgrade npm if 'npm ci' is unsupported

RomainMuller:


eladb:
We could upgrade npm in superchain

PR #87, opened 2019-04-04
feat: upgrade superchain to dotnet to 2.2.202

eladb:
Upgrading to .NET Core 2.2.202

I am keeping this as a draft until we get some mileage from our CDK PR builds to make sure it's all clear to upgrade.

### DIFF
--- a/superchain/Dockerfile
+++ b/superchain/Dockerfile
@@ -185,7 +185,7 @@ RUN set -ex \
 
 COPY m2-settings.xml $MAVEN_CONFIG/settings.xml
 
-### .NET CLI -- https://github.com/aws/aws-codebuild-docker-images/blob/master/ubuntu/dot-net/core-2.1/Dockerfile
+### .NET CLI
 
 # Install .NET CLI dependencies
 RUN set -ex \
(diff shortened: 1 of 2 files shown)


superchain/Dockerfile
```diff
@@ -185,7 +185,7 @@ RUN set -ex \
 
 COPY m2-settings.xml $MAVEN_CONFIG/settings.xml
 
-### .NET CLI 2.1 -- https://github.com/aws/aws-codebuild-docker-images/blob/master/ubuntu/dot-net/core-2.1/Dockerfile
+### .NET CLI -- https://github.com/aws/aws-codebuild-docker-images/blob/master/ubuntu/dot-net/core-2.1/Dockerfile
```
RomainMuller: Well so you removed the 2.1 from before the URL, but what about the one... IN the URL?

superchain/README.md
```diff
@@ -4,7 +4,7 @@ This docker image includes the following tools:
 
  - [Node.js 8.11.0](https://nodejs.org/download/release/v8.11.0/)
  - [Java OpenJDK 8](http://openjdk.java.net/install/)
- - [.NET Core 2.0](https://www.microsoft.com/net/download)
+ - [.NET Core 2.2](https://www.microsoft.com/net/download)
```
RomainMuller: Oh good catch! 👌🏻

superchain/Dockerfile
eladb: Actually, there is no codebuild dockerfile for 2.2

PR #90, opened 2019-04-09
 feat: multiple artifacts for GitHub and Shellable 

rix0rrr:
Add multiple output artifact support for Shellable (including an initial
attempt at a BuildSpec helper as requested in awslabs/aws-cdk#2039).

Add multiple input artifact support for the GitHub publisher. The
tag and release notes will be taken from the primary artifact, the other
artifacts will be attached to the same release.

This is in preparation of publishing downloadable reference docs to
GitHub.

### DIFF
--- a/lib/build-spec.ts
+++ b/lib/build-spec.ts
@@ -4,6 +4,22 @@ const MAGIC_ARTIFACT_NAME = 'PRIMARY';
 
 /**
  * Class to model a buildspec version 0.2
+ *
+ * Artifact handling is a little special: CodeBuild will interpret the
+ * 'artifacts' section differently depending on whether there are secondary
+ * artifacts or not.
+ *
+ * If there is only one artifact, the single artifact must go into the top-level
+ * 'artifacts' section. If there are multiple artifacts, all of them must go
+ * into the 'secondary-artifacts' section. Upon rendering to JSON, the caller
+ * must supply the name of the primary artifact (it's determined by
+ * the CodePipeline Action that invokes the CodeBuild Project that uses this
+ * buildspec).
+ *
+ * INVARIANT: in-memory, the BuildSpec will treat all artifacts the same (as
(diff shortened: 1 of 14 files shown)


lib/build-spec.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  public render(): BuildSpecStruct {
+    return this.spec;
+  }
+}
+
+export interface MergeOptions {
+  /**
+   * Rename the primary artifact on the right-hand-side of the merge operation, if present
```
eladb: Mention that this is required if the rhs config includes a primary artifact

lib/shellable.ts
```diff
@@ -145,23 +146,6 @@ export interface AssumeRole {
   externalId?: string;
 }
 
-/**
- * Properties used to create a Shellable
- */
-export interface ShellableProps extends ShellableOptions {
```
eladb: What’s up here? No props?

rix0rrr: Duplicate definition.

eladb: Not sure what that means? What am I missing?

rix0rrr: We had 2 definitions of this interface in the same file. They were both exactly the same.

eladb: 👍

lib/build-spec.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+export interface SimpleBuildSpecProps {
+  preBuild?: string[];
+  build?: string[];
+
+  /**
+   * Where the directories for each artifact are
+   *
```
eladb: I kind of liked the previous API better that optimized for a single primary artifact:

```ts
interface SimpleBuildSpecProps {
  // ...
  artifactsDir: string;

  additionalArtifactDirs? : { [ ... ] }
}

PR #91, opened 2019-04-10
feat: support npm disttags

eladb:
When publishing to npm, you can now specify `distTag: 'TAG'`
and we will publish with `--tag=TAG`.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/publishing.ts
+++ b/lib/publishing.ts
@@ -83,6 +83,13 @@ export interface PublishToNpmProjectProps {
    * @default true
    */
   dryRun?: boolean;
+
+  /**
+   * npm dist-tag to use when publishing artifacts.
+   *
+   * @default - npm default behavior ("latest" unless dist tag is specified in package.json)
+   */
+  distTag?: string;
 }
 
 /**
@@ -103,7 +110,8 @@ export class PublishToNpmProject extends cdk.Construct implements IPublisher {
       entrypoint: 'publish.sh',
       environment: {
         FOR_REAL: forReal,
-        NPM_TOKEN_SECRET: props.npmTokenSecret.secretArn
+        NPM_TOKEN_SECRET: props.npmTokenSecret.secretArn,
+        DISTTAG: props.distTag || ''
       },
     });
 
--- a/lib/publishing/npm/publish-npm.sh
+++ b/lib/publishing/npm/publish-npm.sh
@@ -51,7 +51,7 @@ for TGZ in $(find ${PWD}/js -iname '*.tgz'); do
     # returns an empty string if the package exists, but version doesn't
     npm_view=$(npm view ${mod}@${ver} 2> /dev/null || true)
     if [ -z "${npm_view}" ]; then
-        $dry_npm publish $TGZ --access=public "${DISTTAG}" --loglevel=silly
+        $dry_npm publish $TGZ --access=public ${DISTTAG} --loglevel=silly
     else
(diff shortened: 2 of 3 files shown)


lib/publishing.ts
```diff
@@ -103,7 +110,8 @@ export class PublishToNpmProject extends cdk.Construct implements IPublisher {
       entrypoint: 'publish.sh',
       environment: {
         FOR_REAL: forReal,
-        NPM_TOKEN_SECRET: props.npmTokenSecret.secretArn
+        NPM_TOKEN_SECRET: props.npmTokenSecret.secretArn,
+        DISTTAG: props.distTag || ''
```
RomainMuller: What's wrong in leaving `undefined` here?

lib/publishing/npm/publish-npm.sh
```diff
@@ -22,6 +22,11 @@ fi
 # NPM #
 #######
 
+DISTTAG=${DISTTAG:-""}
```
RomainMuller: The `""` is unnecessary here.

lib/publishing/npm/publish-npm.sh
```diff
@@ -22,6 +22,11 @@ fi
...
+if [ -n "${DISTTAG}" ]; then
+    DISTTAG="--tag=${DISTTAG}"
```
RomainMuller: I find it confusing that the semantics of the `DISTTAG` variable changes at this point.

lib/publishing/npm/publish-npm.sh
```diff
@@ -40,13 +45,13 @@ for TGZ in $(find ${PWD}/js -iname '*.tgz'); do
     ver="$(node -e "console.log(${packageInfo}.version);")"
 
     echo "-------------------------------------------------------------------------------------------------"
-    echo "Publishing to npm: ${mod}@${ver} from $TGZ"
+    echo "Publishing to npm: ${mod}@${ver} from $TGZ ${DISTTAG}"
```
RomainMuller: Would put the `$DISTTAG` before the `$TGZ`, so it's closer to the version.

test/test-stack.ts
```diff
@@ -88,7 +88,8 @@ export class TestStack extends cdk.Stack {
     //
 
     pipeline.publishToNpm({
-      npmTokenSecret: { secretArn: 'arn:aws:secretsmanager:us-east-1:712950704752:secret:delivlib/npm-OynG62' }
+      npmTokenSecret: { secretArn: 'arn:aws:secretsmanager:us-east-1:712950704752:secret:delivlib/npm-OynG62' },
+      // distTag: 'next'
```
RomainMuller: Why is this commented?

lib/publishing.ts
eladb: The way `environment` is defined right now, the value cannot be undefined.

lib/publishing/npm/publish-npm.sh
eladb: I know, but I wanted to be explicit about.

lib/publishing/npm/publish-npm.sh
eladb: It's a matter of style and I rather not have to run this test again (it takes like 30 minutes).

test/test-stack.ts
eladb: It's just an example, I can remove.

PR #96, opened 2019-06-03
fix: Stop pulling GPG keys from the internets

RomainMuller:
It has been proven that the internets cannot be trusted with this task 🤷🏻‍♂️


eladb:
I believe there is another instance of downloading gpg keys in the Dockerfile.

RomainMuller:
The GPG file has been verified "complete" as follows (pardon the French):

```console
$ gpg --dry-run --import < node-release-keys.asc
gpg: key D7062848A1AB005C: 1 signature not checked due to a missing key
gpg: key 7434390BDBE9B9C5: 5 signatures not checked due to missing keys
gpg: key B63B535A4C206CA9: 9 signatures not checked due to missing keys
gpg: key B01FBB92821C587A: 1 signature not checked due to a missing key
gpg: key C97EC7A07EDE3FC1: 23 signatures not checked due to missing keys
gpg: key 09FE44734EB7990E: 5 signatures not checked due to missing keys
gpg: key 770F7A9A5AE15600: 1 signature not checked due to a missing key
gpg: key E73BC641CC11F4C8: 14 signatures not checked due to missing keys
gpg: key C273792F7D83545D: 7 signatures not checked due to missing keys
gpg: key F07496B3EB3C1762: 1 signature not checked due to a missing key
gpg:       Quantité totale traitée : 11
```

rix0rrr:
Might I suggest that a less drastic solution would be to move the "download keys" command to the very top of the Dockerfile, so it can be reused?

RomainMuller:
@rix0rrr - that will not make the key servers less prone to failure. Those have been notoriously brittle over the time I've monitored them. I also find it is a tighter lock having the keys checked in (pinned in source) rather than relying on a remote resource, which is susceptible to being hijacked in unpredictable ways (this means someone would have found a hash collision on one or more of those keys... which is unlikely... but still).

PR #133, opened 2019-08-26
feat: use mono's signcode to sign .NET assemblies

RomainMuller:
Now that Mono 6 is out we switched to the `jsii/superchain` image, which
uses this release, we are able to use `signcode` to issue `sha256`
signatures, instead of using un-maintained `osslsigncode`.

---


rix0rrr:
Tested?

RomainMuller:
Hah! Turns out the publisher's not using the `jsii/superchain` image. Well starting another test w/ a fixed up version... 🤷🏻‍♂️

PR #153, opened 2019-09-30
fix(nuget): migrate to snupkg symbols package format

RomainMuller:
For more info: https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg

-----


lib/publishing/nuget/publish.sh
```diff
@@ -62,7 +61,11 @@ for NUGET_PACKAGE_PATH in $(find dotnet -name *.nupkg -not -iname *.symbols.nupk
         fi
     fi
     echo "📦  Publishing ${NUGET_PACKAGE_PATH} to NuGet"
-    dotnet nuget push $NUGET_PACKAGE_PATH -k $NUGET_API_KEY -s $NUGET_SOURCE -ss $NUGET_SYMBOL_SOURCE | tee ${log}
+    (
```
rix0rrr: Maybe do an 'if' for old symbols file vs new symbols file and publish to the appropriate server? Or is that not how it works?

PR #184, opened 2019-11-09
feat: migrate library to General Availability CDK version

skinny85:
This updates the CDK libraries to a post-1.0.0 version. It contains only the minimal amount of changes needed for the code to compile & pass tests. Further changes, needed to accommodate the changes in CDK conventions that happened since this library was created, will be coming later.

BREAKING CHANGE:
* AutoBuildOptions.buildSpec is now of type codebuild.BuildSpec
* CanaryProps.scheduleExpression is now of type events.Schedule, and was renamed to 'schedule'
* ChangeControllerProps.scheduleExpression is now of type events.Schedule, and was renamed to 'schedule'
* RsaPrivateKeySecretProps.deletionPolicy has been renamed to 'removalPolicy'
* PipelineProps.buildSpec is now of type codebuild.BuildSpec
* GitHubRepo.tokenParameterName is now of type cdk.SecretValue, and was renamed to 'token'
* ShellableOptions.alarmPeriodSec is now of type cdk.Duration, and was renamed to 'alarmPeriod'

-----


custom-resource-handlers/src/pgp-secret.ts
```diff
@@ -153,3 +153,11 @@ async function _getPublicKey(secretArn: string): Promise<string> {
     await _rmrf(tempDir);
   }
 }
+
+async function _deleteSecret(event: cfn.DeleteEvent): Promise<cfn.ResourceAttributes> {
```
eladb: I believe @RomainMuller intentionally omitted this (although not sure why)

lib/auto-build.ts
```diff
@@ -43,19 +43,10 @@ export class AutoBuild extends Construct {
     const project = new codebuild.Project(this, 'Project', {
       source: props.repo.createBuildSource(this, true),
       environment: createBuildEnvironment(props.environment),
-      badge: true,
+      badge: props.repo.allowsBadge,
       buildSpec: props.buildSpec
     });
 
-    // not support in this version of the cdk
-    const cfnProject = project.node.tryFindChild('Resource') as Resource;
```
eladb: Is this a regression?

rhboyd:
This is a great update. I can’t wait for us to be able to share it with the community. I’m going to have to take out a second mortgage to afford all the beers I owe @skinny85 at re:Invent.

custom-resource-handlers/src/pgp-secret.ts
skinny85: The reason I added it was because without it, we need to manually delete the secret between the runs of our integration tests (`npm run test update`) using the AWS CLI (the console does not allow immediately deleting a secret). Note that the default for this resource is `RemovalPolicy.RETAIN`.

If we need it, I think it's reasonable to expect that other customers of `delivlib` will need it too.

@RomainMuller are you OK with adding this delete capability to `OpenPGPKeyPair`?

lib/auto-build.ts
skinny85: No. Basically, this is now supported in the CDK directly, so no need to use an L1 for it.

What is more, webhooks are not allowed to be used with CodeCommit as the source of a CodeBuild project, so we shouldn't be adding this unconditionally - instead, we should rely on `source: props.repo.createBuildSource(this, true),` to make the decision whether the webhook can be created or not.

custom-resource-handlers/src/pgp-secret.ts
RomainMuller: I had left it out because:
- You should NOT take the private key out of the AWS Account (the point is that no human ever has access to that)
- This implies you may not have a backup of it outside of the cloud
- If you inadvertently lose it for whatever reason, you might be in a very difficult spot.

If we implement `Delete` actions, we __MUST__ not force immediate deletion (or at least force the customer to very explicitly opt into that behavior). It's way too risky.

skinny85: How about something like this?

```ts
interface OpenPGPKeyPairProps {
 // other properties as today...

  /** @default OpenPGPKeyRemovalPolicy.RETAIN */
  readonly removalPolicy?: OpenPGPKeyRemovalPolicy;
}

export enum OpenPGPKeyRemovalPolicy {
  RETAIN, DESTROY_SAFELY, DESTROY_IMMEDIATELY
}
```
```

Comment (RomainMuller) [reply custom-resource-handlers/src/pgp-secret.ts]: I'd be okay with that.

Comment (skinny85) [reply custom-resource-handlers/src/pgp-secret.ts]: Done in the newest revision.

**Comment by skinny85** [comment]
Published a new version with the changes to `OpenPGPKeyPair` @RomainMuller approved in a comment.

**Comment by skinny85** [comment]
Ah. Forgot `watch` doesn't build the custom resource lambdas... classic mistake. Will update in a second.

**Comment by skinny85** [comment]
Fixed the asset hashes.

File: lib/open-pgp-key-pair.ts [hunk 3]
```diff
@@ -81,39 +115,46 @@ export class OpenPGPKeyPair extends cdk.Construct implements ICredentialPair {
       description: 'Generates an OpenPGP Key and stores the private key in Secrets Manager and the public key in an SSM Parameter',
       code: new lambda.AssetCode(codeLocation),
       handler: 'index.handler',
-      timeout: 300,
-      runtime: lambda.Runtime.NodeJS810,
+      timeout: cdk.Duration.seconds(300),
+      runtime: lambda.Runtime.NODEJS_8_10,
```
eladb: Should we use 10.x?

eladb:
Any updates here?

lib/open-pgp-key-pair.ts
skinny85: I'd rather leave this as-is, if you don't mind. I think this should be a separate change, and we still have some time to use Node 8 in Lambda.

skinny85:
> Probably needs a major version bump of delivlib since APIs are broken.

Yes. I believe the automatic release script would take care of that, thanks for conventional commits.

However, I've disabled the release pipeline for now. I don't think we should have another release until we've made the API more aligned with CDK post-GA version conventions. I was going to do that in a subsequent PR, after merging this (I didn't want to make this gigantic PR even bigger).

eladb:
@skinny85  what's up with this? Let's land this.

skinny85:
> @skinny85 what's up with this? Let's land this.

I'm ready when you are...

richardhboyd:
Landing this would be a huge favor to me. I was planning on talking about publishing L3 constructs in my DEM131 demo.

PR #226, opened 2020-01-21
feat: upgrade Custom Resources to use Lambda's Node10 runtime

### LINKED ISSUE #220: Upgrade lambda runtime to use NODEJS_10 instead of 8

skinny85:
Our custom resources for signing and key management shell out to OpenSSL and GPG programs. Those binaries no longer ship with Lambda's Node runtimes later than 8. This changes the Lambda functions backing the custom resources to include layers that add those binaries. In the case of OpenSSL, just the CLI binary is enough, as the libraries are present in the runtime; for GPG, however, we also need to include a bunch of shared libraries, and the `gpg-agent` binary, required for managing private keys with GPG 2.0.

All binaries were obtained by downloading them from an EC2 instance running Amazon Linux 2,
and zipping them up to be used in a layer.

The invocation of the commands had to change to refer to the binaries in `/opt/`, which is where the Lambda layers wind up, and some additional required options were added because of the way Lambda gives read-only permissions to the underlying filesystem.

Fixes #220

-----


skinny85:
> How were these layers created? Do we have a way to update these dependencies?

They were created by downloading binaries from an EC2 instance running Amazon Linux 2.

What we can do, is put the binaries in the repo directly, and let CDK assets do the zipping. It will make the binaries kept in the Git repo larger, but that would make it easier to update them in the future.

skinny85:
@eladb ping

skinny85:
> > How were these layers created? Do we have a way to update these dependencies?
> 
> They were created by downloading binaries from an EC2 instance running Amazon Linux 2.
> 
> What we can do, is put the binaries in the repo directly, and let CDK assets do the zipping. It will make the binaries kept in the Git repo larger, but that would make it easier to update them in the future.

On second thought, you can always just unzip the current files, add stuff to them, and zip them up again, so it probably doesn't make sense to ship a larger package to customers of delivlib.

PR #240, opened 2020-02-12
fix(chime): make Chime notifier actually deploy

rix0rrr:
- The source was too big for inline code and so it didn't actually
  deploy: strip out the source map.
- Move the configuration into the Lambda trigger's input (instead of
  environment variables) so we can use a SingletonFunction for all
  pipelines in the project.
- Add missing permissions to list the pipeline's actions.
- Fix the pattern

Add to Delivlib's own pipeline to properly test it.

-----

### DIFF
--- a/lib/chime-notifier/chime-notifier.ts
+++ b/lib/chime-notifier/chime-notifier.ts
@@ -2,6 +2,7 @@ import { Construct, Duration } from "@aws-cdk/core";
 import * as cpipeline from '@aws-cdk/aws-codepipeline';
 import * as iam from '@aws-cdk/aws-iam';
 import * as lambda from '@aws-cdk/aws-lambda';
+import * as events from '@aws-cdk/aws-events';
 import * as events_targets from '@aws-cdk/aws-events-targets';
 import fs = require('fs');
 import path = require('path');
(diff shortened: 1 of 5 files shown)


lib/chime-notifier/chime-notifier.ts
```diff
@@ -40,33 +43,49 @@ export class ChimeNotifier extends Construct {
...
 
     if (props.webhookUrls.length > 0) {
-      const notifierLambda = new lambda.Function(this, 'Default', {
+      // Reuse the same Lambda code for all pipelines, we will move the Lambda parameterizations into
+      // the CloudWatch Event Input.
+      const notifierLambda = new lambda.SingletonFunction(this, 'Default', {
         handler: 'index.handler',
-        code: lambda.Code.inline(fs.readFileSync(path.join(__dirname, 'notifier-handler.js')).toString('utf8')),
+        uuid: '0f4a3ee0-692e-4249-932f-a46a833886d8',
+        code: lambda.Code.inline(stripComments(fs.readFileSync(path.join(__dirname, 'notifier-handler.js')).toString('utf8'))),
```
eladb: why not just use an asset?

rix0rrr: Kinda just mirrored this off of the pipeline watcher which also uses inline code.

PR #247, opened 2020-02-18
feat(bump): create pull request for bumps

eladb:
Add support for automatically creating pull requests for bumps.

This requires the GitHub token to be available during bump. However, currently we model the GitHub token as a SecretValue, which resolves during deployment (by CloudFormation). This is not secure because it will pass the secret itself as environment variables. To address this, we changed `token` to `tokenSecretArn` and resolve it as needed (either during deployment if it’s required for a CFN definition or during runtime by extracting it from secrets manager).

BREAKING CHANGE: `repo.token` is now `tokenSecretArn` to enforce that the token is kept in AWS SecretsManager. Also, the API for `IRepo.createBuildSource` was modified.

### DIFF
--- a/lib/bump/bump.ts
+++ b/lib/bump/bump.ts
@@ -21,6 +21,12 @@ export interface AutoBumpOptions {
   bumpCommand?: string;
 
   /**
+   * The command to determine the current version.
+   * @default "git describe" by default the latest git tag will be used to determine the current version
+   */
+  versionCommand?: string;
+
+  /**
    * The schedule to produce an automatic bump.
    *
    * The expression can be one of:
@@ -71,9 +77,23 @@ export interface AutoBumpOptions {
    * The name of the branch to push the bump commit (e.g. "master")
    * This branch has to exist.
    *
-   * @default the commit will be pushed to the branch `bump/$(git describe)`
+   * @default the commit will be pushed to the branch `bump/$VERSION`
    */
   branch?: string;
+
+  /**
+   * Create a pull request after the branch is pushed.
+   *
+   * @default false
+   */
+  pullRequest?: boolean;
+
+  /**
+   * Options for pull request
+   *
+   * @default - default options
+   */
+  pullRequestOptions?: PullRequestOptions;
 }
 
 export interface AutoBumpProps extends AutoBumpOptions {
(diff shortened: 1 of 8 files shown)


lib/bump/bump.ts
```diff
@@ -83,6 +110,29 @@ export interface AutoBumpProps extends AutoBumpOptions {
...
+   *
+   * @default "chore(release): $VERSION"
+   */
+  title?: string;
+
+  /**
+   * The PR body.
+   * @default "see CHANGELOG"
+   */
+  body?: string;
```
rix0rrr: Given that the most important part of the human verification here would be verifying the CHANGELOG, I would prefer to see it pasted directly in the PR body so we can see at a glance whether it is formatted correctly and doesn't contain strange items.

lib/bump/bump.ts
```diff
@@ -187,3 +253,26 @@ export class AutoBump extends cdk.Construct {
...
+    base: options.base ?? `master`,
+    head: `$BRANCH`
+  };
+
+  const curl = [
+    `curl`,
+    `-X POST`,
+    `--header "Authorization: token $GITHUB_TOKEN"`,
+    `--header "Content-Type: application/json"`,
+    `-d ${JSON.stringify(JSON.stringify(request))}`, // to escape quotes
```
rix0rrr: lol

lib/bump/bump.ts
```diff
@@ -71,9 +77,30 @@ export interface AutoBumpOptions {
    * The name of the branch to push the bump commit (e.g. "master")
    * This branch has to exist.
```
rix0rrr: This can't be true if we're making a PR.

test/expected.yml
```diff
@@ -466,7 +466,7 @@ Resources:
                 Owner: awslabs
                 Repo: aws-delivlib-sample
                 Branch: master
-                OAuthToken: "{{resolve:secretsmanager:github-token:SecretString:::}}"
+                OAuthToken: "{{resolve:secretsmanager:arn:aws:secretsmanager:us-east-1:712950704752:secret:github-token-nnAqfW:SecretString:::}}"
```
rix0rrr: Um. Is this correct?

rix0rrr:
In your PR message, it will be more helpful to users if the `BREAKING CHANGE` section only focuses on code changes they need to make. Rationale is good and all but I'm not sure people skimming the change log are ready to deal with a big blocks of text. 

Rationale should be included in the PR body, just not in the `BREAKING CHANGE` block.

Also, you missed the function signature change of `createBuildSource`.

lib/bump/bump.ts
eladb: Works...

lib/bump/bump.ts
eladb: I'll add a little check that forbids this case.

lib/bump/bump.ts
eladb: I will leave it for a subsequent PR. It might be a bit tricky to support this at the delivlib level because we are not opinionated about changelogs here.

test/expected.yml
eladb: I think it is, but let me double check

eladb: Yes, it can be either the name or the ARN of the secret (from [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#dynamic-references-secretsmanager)).

eladb:
@rix0rrr 

- Updated PR message
- Added check that head!=base if PR is enabled
- Added a unit test for bumps

PR #251, opened 2020-02-24
feat: Support Env Var Types in Shellable

MrArnoldPalmer:
Adds support for different environment variable types in scriptable
environment. This allows referencing ssm parameters and secrets manager
values as environment variables.

Bump cdk version dependencies to latest. Had to bump to >=v1.20 for access to `codebuild.BuildEnvironmentVariableType.SECRETS_MANAGER`.

This is needed for accessing our github access token as an environment variable in @jsii/integ-tests module.

-----


lib/shellable.ts
```diff
@@ -34,7 +45,7 @@ export interface ShellableOptions {
    *
    * @default No additional environment variables
    */
-  environment?: { [key: string]: string };
+  environment?: { [key: string]: string | ShellableEnvironmentVariable };
```
eladb: This is not jsii-able (although we don’t build this module with jsii currently). What you could do is just add another field “environmentSecrets” which will allow specifying secret ARNs for values. 

But I am also ok with this.

MrArnoldPalmer: yeah wondering what the possibility of building with JSII is for delivlib. Was just looking for the best way to add this without breaking the api.

I see what you mean with `environmentSecrets`. We would have to do the same for ssm parameters with `environmentParameters` or something similar. I'll explore a bit but first instinct was to mirror the codebuild construct API.

MrArnoldPalmer: yeah wondering what the possibility of building with JSII is for delivlib. Was just looking for the best way to add this without breaking the api.

I see what you mean with `environmentSecrets`. We would have to do the same for ssm parameters with `environmentParameters` or something similar. I'll explore a bit but first instinct was to mirror the codebuild construct API.

MrArnoldPalmer: yeah wondering what the possibility of building with JSII is for delivlib. Was just looking for the best way to add this without breaking the api.

I see what you mean with `environmentSecrets`. We would have to do the same for ssm parameters with `environmentParameters` or something similar. I'll explore a bit but first instinct was to mirror the codebuild construct API.

test/shellable.test.ts
```diff
@@ -118,3 +118,87 @@ test('privileged mode', () => {
...
+                          Ref: "AssetParameters3d34b07ba871989d030649c646b3096ba7c78ca531897bcdb0670774d2f9d3e4S3VersionKeyF3F83F76"
+                        }
+                      ]
+                    }
+                  ]
+                }
+              ]
+            ]
+          }
+        },
```
MrArnoldPalmer: Is there a way to ignore pieces of nested properties in assertions? Then we don't have to care about things we aren't necessarily testing.

MrArnoldPalmer:
pipeline-diff output of note.
```sh
Resources
[~] AWS::KMS::Key GitHubPipeline/BuildPipeline/ArtifactsBucketEncryptionKey GitHubPipelineBuildPipelineArtifactsBucketEncryptionKeyFED13A11
 └─ [~] KeyPolicy
     └─ [~] .Statement:
         └─ @@ -13,7 +13,9 @@
            [ ]   "kms:Delete*",
            [ ]   "kms:ScheduleKeyDeletion",
            [ ]   "kms:CancelKeyDeletion",
            [-]   "kms:GenerateDataKey"
            [+]   "kms:GenerateDataKey",
            [+]   "kms:TagResource",
            [+]   "kms:UntagResource"
            [ ] ],
            [ ] "Effect": "Allow",
            [ ] "Principal": {
[~] AWS::S3::Bucket GitHubPipeline/BuildPipeline/ArtifactsBucket GitHubPipelineBuildPipelineArtifactsBucket60481C14
 └─ [+] PublicAccessBlockConfiguration
     └─ {"BlockPublicAcls":true,"BlockPublicPolicy":true,"IgnorePublicAcls":true,"RestrictPublicBuckets":true}
[~] AWS::Lambda::Function SingletonLambda0f4a3ee0692e4249932fa46a833886d8 SingletonLambda0f4a3ee0692e4249932fa46a833886d8CE9FA488
 └─ [~] Code
     └─ [~] .ZipFile:
                   .....
```

test/shellable.test.ts
skinny85: Yes - you can use `haveResourceLike` to match only a part of the resulting template.

PR #253, opened 2020-02-27
fix: Grant env var access in shellable when needed

MrArnoldPalmer:
Calls `grantRead` methods for secrets and SSM parameters passed as env
vars to shellable. Originally meant for consumers to handle on their end
but constructs don't necessarily expose the roles to be given access.

-----

### DIFF
--- a/lib/shellable.ts
+++ b/lib/shellable.ts
@@ -4,6 +4,8 @@ import cpipeline = require('@aws-cdk/aws-codepipeline');
 import cpipeline_actions = require('@aws-cdk/aws-codepipeline-actions');
 import iam = require('@aws-cdk/aws-iam');
 import assets = require('@aws-cdk/aws-s3-assets');
+import { Secret } from '@aws-cdk/aws-secretsmanager';
+import { StringParameter } from '@aws-cdk/aws-ssm';
 import cdk = require('@aws-cdk/core');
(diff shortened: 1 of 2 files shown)


lib/shellable.ts
```diff
@@ -247,6 +249,18 @@ export class Shellable extends cdk.Construct {
     this.role = this.project.role!; // not undefined, as it's a new Project
     asset.grantRead(this.role);
 
+    // Grant read access to secrets
+    Object.entries(props.environmentSecrets ?? {}).forEach(([name, secretArn]) => {
+      const secret = Secret.fromSecretArn(this, `${name}Secret`, secretArn);
```
shivlaks: question: I'm guessing we have a convention where we add the suffix of `Secret` and `Parameter` accordingly?
Is there always going to be an `ISecret` / what happens when `fromSecretArn` can't find a Secret?

do we need to validate anything before adding the grantRead?

MrArnoldPalmer: No convention, just needed indentifiers here. We could use the name as is I suppose.

Whats the recommended way for verifying that a `X.fromXyz()` call resolved to something?

shivlaks: I'm not sure actually, but I think what you have is fine based on the examples i see in our codebase

PR #259, opened 2020-03-05
chore(pipeline): configure name and expose build project

nija-at:
Exposing the CodeBuild project in the pipeline will enable use cases
around creating monitoring resources. As an example, the build's
Duration and Failure metrics can be graphed on a CloudWatch dashboard.

Having a concrete name to the CodeBuild Project allows for the above
mentioned dashboards to reliably graph these across syntheses.

-----

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -1,8 +1,16 @@
-import { aws_cloudwatch as cloudwatch, aws_codebuild as cbuild,
-  aws_codepipeline as cpipeline, aws_codepipeline_actions as cpipeline_actions,
-  aws_events as events, aws_events_targets as events_targets, aws_iam as iam,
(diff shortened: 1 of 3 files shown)


nija-at:
> In the past I've avoided exposing the underlying resources of delivlib constructs cause it feels we are trying to keep the abstraction relatively tight.

I hear ya. 
The other option I was considering was to expose the build metrics right out of the delivlib's `pipeline.ts` via an API, similar to what [CodeBuild's `Pipeline` construct](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-codebuild.Project.html#metric-wbr-buildsprops) does. Do you think that would be better approach? Any other ideas?

MrArnoldPalmer:
I do think this is the right approach (exposing properties with underlying constructs) personally. Just wanted to spur any conversation that's worth having about it. We can see how this shakes out from the consumer side.

PR #324, opened 2020-05-03
refactor(bump): refactor AutoBump to support more use-case more cleanly

iliapolo:
The current `AutoBump` construct has outgrown its original intent.

For example:

- It defaults `scheduleExpression` to `cron(0 12 * * ? *)`. While this was ok for `delivlib` itself, if forces us to use a magic value of `disable` in aws-cdk since we manually trigger bump builds.

- It defaults `bumpCommand` to `/bin/bash ./bump.sh`. While this is mostly ok, when we added the merge-back PR creation, since it doesn't actually bump anything, if forced us to again [introduce](https://github.com/awslabs/aws-delivlib/blob/master/lib/bump/bump.ts#L254) this magic `disable` value.

In addition, the code seemed rather cluttered and hard to follow. Hopefully this PR will rectify that.

#### `AutoPullRequest` construct  

Extract most of the logic from `AutoBump` to a generic `AutoPullRequest` construct that exposes all the necessary features to implement various scenarios, for example:

- `commands`: Which commands to run on the `head` branch.

  > This means we can leave it blank for merge-back, and use `/bin/bash ./bump.sh` for the bump.

- `exports`: Key value mapping for exporting env variables that will be accessible after the command have executed.

  > For example: `VERSION: node -p "require('./scripts/get-version')"` will make the `VERSION` env variable available and its value will be the output of the command.   

The existing `AutoBump` and `AutoMergeBack` constructs simply use `AutoPullRequest` and provide some relevant defaults.

#### Monocdk version upgrade

Monocdk version was upgraded to 1.49.0. This required changing a bunch of imports, which is why there are so many files changed in the PR. I'm not a 100% sure I did the imports correctly, would appreciate a validation.

#### Merge Back Stage

The `AutoMergeBack` build can now be configured as part of the pipeline, so we won't need to manually trigger it, it will simply by triggered after the Publish stage:

```typescript
releasePipeline.autoMergeBack({
  versionCommand: `node -p "require('./scripts/get-version')"`,
  labels: ['pr/no-squash'],
  base: {
    name: 'master'
  },
  head: {
    source: 'release'
  },
  stage: {
    name: 'MergeBack',
    after: 'Publish'
  }
})
``` 
#### Breakages 

I've also taken the liberty to change some of the existing API's in favor of reducing code duplication and clarity. So this release will trigger a major version bump.

BREAKING CHANGE: `AutoBump` API has significantly changed. Refer to the docstrings for guidance. 

-----


iliapolo:
To reviewers. Im still working out the integration tests and the pipeline deployment, but would appreciate a look at the code in the meantime. 

Thanks!

lib/pipeline.ts
```diff
@@ -153,13 +154,48 @@ export interface PipelineProps {
...
+   *
+   * @default - The CodeBuild project will be created indepdent of any stage.
+   */
+  readonly stage?: MergeBackStage
+}
+
+/**
+ * Options for configuring an auto bump for this pipeline.
+ */
+export interface AutoBumpOptions extends Omit<AutoBumpProps, 'repo'> {
```
eladb, line 188: how does jsii handle this?

lib/pipeline.ts
```diff
@@ -362,10 +398,39 @@ export class Pipeline extends cdk.Construct {
...
-    return new AutoBump(this, 'AutoBump', {
+    const autoBump = new AutoBump(this, 'AutoBump', {
+      repo: this.repo,
+      ...options
+    });
+
+    return autoBump;
+  }
+
+  public autoMergeBack(options?: AutoMergeBackOptions) {
```
eladb, line 409: docstrings

lib/pipeline.ts
iliapolo: Well, delivlib isn't a `jsii` module...

lib/pipeline.ts
iliapolo: 👍

lib/pipeline.ts
nija-at: To answer @eladb's direct question - https://github.com/aws/jsii/issues/1707

nija-at: > Well, delivlib isn't a jsii module...

Nevertheless, this feels very backward.
Are we not able to construct these forward, rather than omit backwards?

nija-at: Start by defining `AutoPullRequestOptions`, then define `AutoMergeBackProps` based on that, etc. etc.

Issue #331, opened 2020-05-06
feature: Support recurring events in ChangeController


pnwpedro:
#### Feature Request
It would be nice to be able to define a recurring event in the `change-control.ics` file, such as:
```
BEGIN:VEVENT
RRULE:FREQ=WEEKLY;INTERVAL=1
DTEND:20200511T170000Z
SUMMARY:Outside of business hours (weekend)
DTSTAMP:20200504T163641Z
DTSTART:20200508T220000Z
SEQUENCE:0
END:VEVENT
```

#### Notes
Based on the current implementation, it looks like [time-window.ts](https://github.com/awslabs/aws-delivlib/blob/master/change-control-lambda/lib/time-window.ts) ignores the `rrule` properties that may exist on events. 

We can see `rrule` set [here](https://github.com/jens-maus/node-ical/blob/master/ical.js#L391) in the `node-ical` library by calling `rrule.fromString()`. This means we're getting an instance of RRule, from the [rrule library](https://github.com/jakubroztocil/rrule/blob/master/src/rrule.ts).

We could potentially call [rrule.after(new Date(), true)](https://github.com/jakubroztocil/rrule/blob/master/src/rrule.ts#L226) to get the next event in the recurrence, inclusive of the current date/time.

PR #333, opened 2020-05-07
feat: recurring event support for change controller

pnwpedro:
This change implements this feature request: https://github.com/awslabs/aws-delivlib/issues/331

#### Motivation & Use Case
We use the ChangeController to enable and disable stage transitions in our pipeline based on a calendar of events. This works well when events are singular, however we'd also like to enforce change control on recurring events. For instance, we want to block stage transition into production waves after hours and on weekends.

Sample event for weekends:
```
BEGIN:VEVENT
RRULE:FREQ=WEEKLY;INTERVAL=1
DTEND:20200511T170000Z
SUMMARY:Outside of business hours (weekend)
DTSTAMP:20200504T163641Z
DTSTART:20200508T220000Z
SEQUENCE:0
END:VEVENT
```

-----

### DIFF
--- a/change-control-lambda/lib/time-window.ts
+++ b/change-control-lambda/lib/time-window.ts
@@ -19,6 +19,8 @@ export interface CalendarEvent {
   params?: any[];
   /** The type of the boundaries for the event */
   datetype: 'date-time';
+  /** A recurrence rule for the event. */
+  rrule?: any;
 }
 type Events = { [uuid: string]: CalendarEvent };
 
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
   return blocks.length > 0 ? blocks[0] : undefined;
 }
 
+/**
+ * Returns the previous and next events for a recurring event surrounding the
+ * provided date. If the date provided is equal to the start of an event, 
+ * the event for that date and the following event will be returend
+ * 
+ * If the provided event is not recurring, no events are returned.
+ *
+ * @param event a recurring calendar event.
+ * @param date the date for which the previous and next event should be returned.
+ */
+function getRecurringEvents(event: CalendarEvent, date: Date): CalendarEvent[] {
+  const recurrences: CalendarEvent[] = [];
+  if (event.rrule == null) return recurrences;
+  const duration = (new Date(event.end).getTime() - new Date(event.start).getTime())
+
+  const after = event.rrule.after(date, false);
+  if (after != null) {
+    recurrences.push(buildEventForDuration(new Date(after), duration, event.summary));
+  }
+
+  const before = event.rrule.before(date, true);
+  if (before != null) {
+    recurrences.push(buildEventForDuration(new Date(before), duration, event.summary));
+  }
+  return recurrences;
+}
+
+/**
+ * A function to build a CalendarEvent given a start date and a duration.
+ *
+ * @param start a start date for the event
+ * @param duration a duration for the event in milliseconds
+ * @param summary a summary to apply to the event
+ */
+function buildEventForDuration(start: Date, duration: number, summary: string): CalendarEvent {
+  const end = new Date(start.getTime() + duration);
+  return {
+    summary: `${summary} ${start.toISOString()} - ${end.toISOString()}`,
+    start: start,
+    end: end,
+    datetype: 'date-time',
+    type: 'VEVENT'
+  };
+}
+
 function containingEventsWithMargin(events: Events, date: Date, advanceMarginSec: number): CalendarEvent[] {
   const bufferedDate = new Date(date.getTime() + advanceMarginSec * 1_000);
+
   return Object.values(events)
     .filter(e => e.type === 'VEVENT')
+    .reduce((arr, e) => {
+      if (e.rrule != null) {
+        // Turn a recurrence rule into the events starting on or before
+        // the date, and the next event starting after the date.
+        arr.push(...getRecurringEvents(e, date));
+      } else {
+        arr.push(e);
+      }
+      return arr;
+    }, [] as CalendarEvent[])
     .filter(e => overlaps(e, { start: date, end: bufferedDate }));
 }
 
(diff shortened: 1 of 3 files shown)


change-control-lambda/lib/time-window.ts
```diff
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
...
+ * Returns the previous and next events for a recurring event surrounding the
+ * provided date. If the date provided is equal to the start of an event, 
+ * the event for that date and the following event will be returend
+ * 
+ * If the provided event is not recurring, no events are returned.
+ *
+ * @param event a recurring calendar event.
+ * @param date the date for which the previous and next event should be returned.
+ */
+function getRecurringEvents(event: CalendarEvent, date: Date): CalendarEvent[] {
```
nija-at, line 55: Better naming: `flattenRecurringEvent()`

change-control-lambda/lib/time-window.ts
```diff
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
...
+  };
+}
+
 function containingEventsWithMargin(events: Events, date: Date, advanceMarginSec: number): CalendarEvent[] {
   const bufferedDate = new Date(date.getTime() + advanceMarginSec * 1_000);
+
   return Object.values(events)
     .filter(e => e.type === 'VEVENT')
+    .reduce((arr, e) => {
+      if (e.rrule != null) {
```
nija-at, line 96:
```suggestion
      if (e.rrule) {
```

change-control-lambda/lib/time-window.ts
```diff
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
...
+ * the event for that date and the following event will be returend
+ * 
+ * If the provided event is not recurring, no events are returned.
+ *
+ * @param event a recurring calendar event.
+ * @param date the date for which the previous and next event should be returned.
+ */
+function getRecurringEvents(event: CalendarEvent, date: Date): CalendarEvent[] {
+  const recurrences: CalendarEvent[] = [];
+  if (event.rrule == null) return recurrences;
```
nija-at, line 57:
```suggestion
  if (!event.rrule) return recurrences;
```

change-control-lambda/lib/time-window.ts
```diff
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
...
+  const duration = (new Date(event.end).getTime() - new Date(event.start).getTime())
+
+  const after = event.rrule.after(date, false);
```
nija-at, line 60: This tells me that the type of `rrule` is not `any`. What's the correct type that should be assigned here?

change-control-lambda/test/time-window.test.ts
```diff
@@ -30,6 +30,15 @@ DTEND:20190204T075900Z
...
 END:VEVENT
 
+BEGIN:VEVENT
+RRULE:FREQ=WEEKLY;INTERVAL=1
+DTEND:20200504T170000Z
+SUMMARY:Block4
+DTSTAMP:20200501T163641Z
+DTSTART:20200501T220000Z
+SEQUENCE:0
+END:VEVENT
```
nija-at, line 40: Would prefer if you could create a separate ics input data for these tests so that it's easier to reason when reading and reasoning about the test post-hoc.

Along with that also add comments what this recurrence in plain english (I'm not familiar with the ics format).

change-control-lambda/test/time-window.test.ts
```diff
@@ -58,3 +67,28 @@ test('a blocked window starts AND finishes within margin', () => {
   const x = shouldBlockPipeline(ics, new Date('2019-02-03T07:00:00.000Z'), 72 * 3_600);
   expect(x && x.summary).toBe('Block3');
 });
+
+test('left edge for recurring event after the initial event', () => {
```
nija-at, line 71: I'm not quite following these test descriptor phrases, and what test cases are being covered.

Can you better phrase, or add any additional comments? What's the initial event and the recurring event here? 
Phrases such as 'current time aligns with the left edge of the first occurrence' might be clearer to understand.

change-control-lambda/lib/time-window.ts
```diff
@@ -40,10 +42,66 @@ export function shouldBlockPipeline(icalData: string | Buffer, now = new Date(),
...
   return Object.values(events)
     .filter(e => e.type === 'VEVENT')
+    .reduce((arr, e) => {
+      if (e.rrule != null) {
+        // Turn a recurrence rule into the events starting on or before
+        // the date, and the next event starting after the date.
+        arr.push(...getRecurringEvents(e, date));
+      } else {
+        arr.push(e);
+      }
```
nija-at, line 102: Would it make sense to push this logic up into `getRecurringEvents()` where it returns the single event in the case it's not recurring. The method can then simply be named `flattenEvents()`

pnwpedro: I think this suggestion makes sense.

change-control-lambda/test/time-window.test.ts
pnwpedro: Cool, makes sense. I'll improve the descriptions/add comments.

change-control-lambda/test/time-window.test.ts
pnwpedro: Will do!

change-control-lambda/lib/time-window.ts
pnwpedro: I'll pull in the type from RRule and add it here.

change-control-lambda/lib/time-window.ts
pnwpedro: I'll change this to flattenEvents() as suggested on the other comment.

PR #359, opened 2020-06-03
feat: rename "env" to "environment" in BuildEnvironmentProps

RomainMuller:
-----

### DIFF
--- a/lib/build-env.ts
+++ b/lib/build-env.ts
@@ -3,7 +3,9 @@ import { aws_codebuild as cbuild } from "monocdk-experiment";
 export interface BuildEnvironmentProps {
   computeType?: cbuild.ComputeType;
   privileged?: boolean;
+  /** @deprecated */
   env?: { [key: string]: string };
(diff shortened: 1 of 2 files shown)


lib/build-env.ts
```diff
@@ -3,15 +3,17 @@ import { aws_codebuild as cbuild } from "monocdk-experiment";
 export interface BuildEnvironmentProps {
   computeType?: cbuild.ComputeType;
   privileged?: boolean;
+  /** @deprecated */
   env?: { [key: string]: string };
+  environment?: { [key: string]: string };
```
eladb, line 8: Why the deprecation? Seems like a similar type...

RomainMuller: What changes is that the callers pass `environment`, not `env`, because everywhere else this is called `environment`.

eladb: okay, so basically this is just a rename. the PR title is quite confusing. I would use `feat: rename "env" to "environment" for BuildEnv`

Issue #358, opened 2020-06-02
Shellable's assumeRole Uses STS Global Endpoint


horsmand:
STS is deprecating the use of the global endpoint and all usage needs to be switched over to the regional endpoints.  The usage of the CLI to call STS to assume a different role in `Shellable` is still using the global endpoint: https://github.com/awslabs/aws-delivlib/blob/master/lib/shellable.ts#L357

I think there are 2 ways this could be fixed, the first being to try and set the Shellable to default to a more recent image than `LinuxBuildImage.UBUNTU_14_04_BASE`. The AWS CLI v2 uses regional endpoints by default, However, I'm not familiar with which version is available on the different images, so I'm not sure if that will work.

The second way to fix it is to set an environment variable in the pre-build steps. Adding `export AWS_STS_REGIONAL_ENDPOINTS=regional` right before the usage of STS through the AWS CLI should work.

PR #377, opened 2020-06-24
feature: expose STS endpoint type

ryyakobe:
AWS STS deprecates the use of the global endpoint and all usage needs to be switched over to the regional endpoints.
As described in this issue: https://github.com/awslabs/aws-delivlib/issues/358, shellable's assumeRole Uses STS Global Endpoint.

This change resolves the issue using the environment variable `AWS_STS_REGIONAL_ENDPOINTS`. 

The default behavior will remain the same as `AWS_STS_REGIONAL_ENDPOINTS` will be set to value `legacy` and the calls will still go to the global endpoint. Although now we expose the boolean flag `useRegionalSTSEndpoints` and if it's set to `true`, then `AWS_STS_REGIONAL_ENDPOINTS=regional` will be used and the calls will go to the regional endpoints.

-----

### DIFF
--- a/lib/shellable.ts
+++ b/lib/shellable.ts
@@ -78,6 +78,14 @@ export interface ShellableOptions {
   buildProjectName?: string;
 
   /**
+   * Indicates if Regional AWS STS endpoints should be used instead
+   * of the global endpoint. Specify true to use Regional AWS STS endpoints.
+   *
+   * @default false
+   */
+  useRegionalStsEndpoints?: boolean;
+
+  /**
    * Can be used to run this build using a specific IAM role. This can be used,
    * for example, to execute in the context of another account (e.g. to run
    * tests in isolation).
@@ -227,7 +235,7 @@ export class Shellable extends cdk.Construct {
     }
 
     this.buildSpec = BuildSpec.simple({
-      preBuild: this.platform.prebuildCommands(props.assumeRole),
+      preBuild: this.platform.prebuildCommands(props.assumeRole, props.useRegionalStsEndpoints),
       build: this.platform.buildCommands(props.entrypoint)
     }).merge(props.buildSpec || BuildSpec.empty());
 
@@ -330,7 +338,7 @@ export abstract class ShellPlatform {
   /**
    * Return commands to download the script bundle
    */
-  public abstract prebuildCommands(assumeRole?: AssumeRole): string[];
+  public abstract prebuildCommands(assumeRole?: AssumeRole, useRegionalStsEndpoints?: boolean): string[];
 
   /**
    * Return commands to start the entrypoint script
@@ -349,7 +357,7 @@ export abstract class ShellPlatform {
 export class LinuxPlatform extends ShellPlatform {
   public readonly platformType = PlatformType.Linux;
 
-  public prebuildCommands(assumeRole?: AssumeRole): string[] {
+  public prebuildCommands(assumeRole?: AssumeRole, useRegionalStsEndpoints?: boolean): string[] {
     const lines = new Array<string>();
     // Better echo the location here; if this fails, the error message only contains
     // the unexpanded variables by default. It might fail if you're running an old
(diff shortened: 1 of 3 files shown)


lib/shellable.ts
```diff
@@ -226,7 +234,7 @@ export class Shellable extends cdk.Construct {
     }
 
     this.buildSpec = BuildSpec.simple({
-      preBuild: this.platform.prebuildCommands(props.assumeRole),
+      preBuild: this.platform.prebuildCommands(props.assumeRole, props.useRegionalSTSEndpoints),
```
rix0rrr, line 237: Why is this not just an environment variable on the CodeBuild project?

ryyakobe:
> 
> 
> Is there a downside to flipping the defaults around, and make regional endpoints the default?
> 
> Are there regions in which it doesn't work yet?
> 
> I get the desire for safety but this seems innocent.

A few quotes from the STS documentation https://docs.aws.amazon.com/STS/latest/APIReference/welcome.html:
> Most AWS Regions are enabled for operations in all AWS services by default. Those Regions are automatically activated for use with AWS STS. Some Regions, such as Asia Pacific (Hong Kong), must be manually enabled.

> Tokens that are valid in all AWS Regions are longer than tokens that are valid in Regions that are enabled by default. Changing this setting might affect existing systems where you temporarily store tokens.

I'm still not sure about all the implications of switching the default to Regional endpoints (this would be a much simpler one line change). So I decided to provide an option for clients, who want to use regional endpoints, but for all the existing code to work as before.

lib/shellable.ts
ryyakobe: Could you please explain in more detail what you mean?
If you mean, that clients that use `aws-delivlib` should set this environment variable by themselves, then I can see 2 problems here:
1. They might just forget to set it.
2. They might not know that `aws-delivlib` makes a call `aws sts assume-role`, so they will not set that variable.

So having it inside of `aws-delivlib` makes it easier to use and less error-prone.

ryyakobe:
Hello! Could someone please take a look at this PR again? Should I simplify it just to
`AWS_STS_REGIONAL_ENDPOINTS=regional aws sts assume-role ...` ?

lib/shellable.ts
ryyakobe: Did you mean adding to [ShellableOptions.environment](https://github.com/awslabs/aws-delivlib/blob/master/lib/shellable.ts#L35) `{ AWS_STS_REGIONAL_ENDPOINTS: regional }`, so it will be used in [cbbuild.Project](https://github.com/awslabs/aws-delivlib/blob/master/lib/shellable.ts#L245) ?

Would it be equivalent to setting this environment variable just before the call assume role in `AWS_STS_REGIONAL_ENDPOINTS=regional aws sts assume-role`?

rix0rrr: It wouldn't be entirely equivalent. The reach would be wider, because it would apply to ALL commands. But that seems like something you'd want.

lib/shellable.ts
```diff
@@ -76,6 +76,14 @@ export interface ShellableOptions {
    */
   buildProjectName?: string;
 
+  /**
+   * Indicates if Regional AWS STS endpoints should be used instead
+   * of the global endpoint. Specify true to use Regional AWS STS endpoints.
+   *
+   * @default false
+   */
+  useRegionalSTSEndpoints?: boolean;
```
rix0rrr, line 85:
```suggestion
  useRegionalStsEndpoints?: boolean;
```

lib/shellable.ts
```diff
@@ -386,8 +395,8 @@ export class LinuxPlatform extends ShellPlatform {
 export class WindowsPlatform extends ShellPlatform {
   public readonly platformType = PlatformType.Windows;
 
-  public prebuildCommands(assumeRole?: AssumeRole): string[] {
-    if (assumeRole) {
+  public prebuildCommands(assumeRole?: AssumeRole, useRegionalSTSEndpoints?: boolean): string[] {
+    if (assumeRole || useRegionalSTSEndpoints) {
```
rix0rrr, line 399: Woah, what? Why are regional endpoints not supported on Windows?

ryyakobe: That was my oversight. I removed the variable from the condition and added undescore to the name as it's now unused.

lib/shellable.ts
ryyakobe: I agree that it's probably something most users would want, but maybe not all.
This small change gives a nice scoped control over this variable and seems to be quite innocent.

lib/shellable.ts
ryyakobe: Renamed this member variable and also a local variable `STSEndpoints` -> `StsEndpoints` in  LinuxPlatform.prebuildCommands.

ryyakobe:
Updated the branch. I will be grateful if someone merges it. Thanks!

PR #431, opened 2020-08-11
fix: make changes to RSA key construct path backwards compatible 

rix0rrr:
The RSA key name used to be calculated off of the signing cert's
complete construct path.

This requires it to be replaced if the construct tree is refactored
(and specifically, if the location of the containing stack is moved,
such as it being moved into a `Stage`).

A name that is based off of the stack's deployment name and the
construct path *inside* the stack:

* is indifferent to the stack's location in the construct tree.
* is backwards compatible with simple setups.

For non-backwards compatible cases, the name is now also configurable.

-----

### DIFF
--- a/test/code-signing-cert.test.ts
+++ b/test/code-signing-cert.test.ts
@@ -41,3 +41,19 @@ test('secret name consists of stack name and relative construct path', () => {
     SecretName: 'ActualStackName/Inbetween/Cert/RSAPrivateKey',
   });
 });
+
+
+test('secret name can be overridden', () => {
+  // WHEN
+  const yetAnotherParent = new cdk.Construct(stack, 'Inbetween');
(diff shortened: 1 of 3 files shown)


test/code-signing-cert.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  commonName: 'CN',
+  country: 'Country',
+  emailAddress: 'Email',
+  locality: 'Locality',
+  organizationName: 'OrgName',
+  organizationalUnitName: 'OrgUnitName',
+  stateOrProvince: 'Province, Please',
+};
+
+test('secret name consists of stack name and relative construct path', () => {
```
RomainMuller, line 29: + a test to confirm the name can be overridden, maybe?

PR #467, opened 2020-09-09
feat: support assume role profiles

iliapolo:
You can now configure a shellable to assume role using named profiles.

-----

### DIFF
--- a/lib/shellable.ts
+++ b/lib/shellable.ts
@@ -208,8 +208,9 @@ export interface AssumeRole {
 
   /**
    * Specify this if you have a long running execution that needs long running sessions.
-   *
    * This will create a profile and use it to delegate credential refreshing to the SDK/CLI
+   *
+   * @default false
    */
   refresh?: boolean;
 
--- a/lib/shellable.ts
+++ b/lib/shellable.ts
@@ -419,14 +419,14 @@ export class LinuxPlatform extends ShellPlatform {
 
     if (assumeRole) {
 
-      const externalId = assumeRole.externalId ? `--external-id "${assumeRole.externalId}"` : '';
-      const StsEndpoints = useRegionalStsEndpoints ? "regional" : "legacy";
-      lines.push(`export AWS_STS_REGIONAL_ENDPOINTS=${StsEndpoints}`);
-
       if (assumeRole.profileName) {
 
         if (!assumeRole.credentialSource && !assumeRole.sourceProfile) {
           throw new Error(`Either 'credentialSource' or 'sourceProfile' is required when using a profile`);
+        }
+
+        if (assumeRole.credentialSource && assumeRole.sourceProfile) {
+          throw new Error(`Only one of 'credentialSource' or 'sourceProfile' is allowed when using a profile`);
         }
 
         const awsHome = assumeRole.awsHomeDir ?? '~/.aws';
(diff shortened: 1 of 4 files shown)


lib/shellable.ts
```diff
@@ -186,6 +186,55 @@ export interface AssumeRole {
...
+  profileName?: string,
+
+  /**
+   * A profile name to be used as the credential with which to assume the role.
+   * Must exist in either the credential or the config file.
+   * Either this or `credentialSource` is required when using `profileName`
+   *
+   * @see https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-source_profile.html
+   */
+  sourceProfile?: string,
```
rix0rrr, line 212: There is no point in this one, really. Yes it's for full capabilities, but in the context of what a Shellable is (just a run-once CodeBuild project) it doesn't make any sense. Who's gonna be writing that other profile entry?

Just drop this.

lib/shellable.ts
```diff
@@ -186,6 +186,55 @@ export interface AssumeRole {
...
+   * @see https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html
+   */
+  profileName?: string,
+
+  /**
+   * A profile name to be used as the credential with which to assume the role.
+   * Must exist in either the credential or the config file.
+   * Either this or `credentialSource` is required when using `profileName`
+   *
+   * @see https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-source_profile.html
```
rix0rrr, line 210: What's the `@default` behavior?

I'd personally still like it if we could just write to the `[default]` profile!

I know that you're triggering the profile-vs-environment variables behavior on the presence of this property.

But how about we add `needsLongSession?: boolean`, trigger the behavior off of that and use `[default]` as the default value for profile if not given?

lib/shellable.ts
```diff
@@ -186,6 +186,55 @@ export interface AssumeRole {
...
+   *
+   */
+  credentialSource?: CredentialSource,
+
+  /**
+   * The directory to create `credentials` and `config` files in case `profileName` is used.
+   *
+   * @default ~/.aws
+   */
+  awsHomeDir?: string;
```
rix0rrr, line 228: This is pointless to configure. There's only one correct location, which is the location the CLI is going to read it from. 

You know where that is, so why even give the user a footgun?

lib/shellable.ts
```diff
@@ -186,6 +186,55 @@ export interface AssumeRole {
...
+  sourceProfile?: string,
+
+  /**
+   * A credential mechanism to use as the credentials with which to assume the role.
+   * Either this or `sourceProfile` is required when using `profileName`
+   *
+   * @see https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-credential_source.html
+   *
+   */
+  credentialSource?: CredentialSource,
```
rix0rrr, line 221: What's the `@default`? Again, given what `Shellable` is honestly the only reasonable thing to put here is `EcsContainer` and nothing else, so why even make the user choose?

lib/shellable.ts
```diff
@@ -369,13 +418,51 @@ export class LinuxPlatform extends ShellPlatform {
...
+        if (assumeRole.credentialSource && assumeRole.sourceProfile) {
+          throw new Error(`Only one of 'credentialSource' or 'sourceProfile' is allowed when using a profile`);
+        }
+
+        const awsHome = assumeRole.awsHomeDir ?? '~/.aws';
+
+        lines.push(`mkdir -p ${awsHome}`);
+        lines.push(`touch ${awsHome}/credentials`);
+        lines.push(`config=${awsHome}/config`);
+        lines.push(`echo [profile ${assumeRole.profileName}]>> $\{config\}`);
```
rix0rrr, line 437: Be aware that it's `[profile foo]` and simply `[default]`.

lib/shellable.ts
```diff
@@ -369,13 +418,51 @@ export class LinuxPlatform extends ShellPlatform {
...
+
+        lines.push(`echo role_session_name = ${assumeRole.sessionName} >> $\{config\}`);
+        lines.push(`echo role_arn = ${assumeRole.roleArn} >> $config`);
+
+        if (assumeRole.externalId) {
+          lines.push(`echo external_id = ${assumeRole.externalId} >> $config`);
+        }
+
+        // let the application code know which role is being used.
+        lines.push(`export AWS_PROFILE=${assumeRole.profileName}`);
```
rix0rrr, line 453: Ah yeah I guess that works too.

rix0rrr: I'm in doubt whether we should set `AWS_SDK_LOAD_CONFIG=1` as well here.

In *principle* that's the JavaScript SDK user's responsibility (AWS CLI doesn't need it for example)... but it feels like a good safeguard against misconfiguration and annoyance downstream.

rix0rrr (changes requested):

lib/shellable.ts
iliapolo: I was thinking about if the shellable is executed in some environment where the shared files actually exist.
Agree its way pre-mature for that.

I'll remove.

nija-at (commented):
Rico is on top of this.

I just wanted to say - good job on slicing this into a PR that's clean to review

lib/shellable.ts
iliapolo: I have a reasoning for that as well, but ok :)

rix0rrr (approved):

PR #544, opened 2020-11-02
feat(pipeline): notify failure on slack

nija-at:
The pipeline now optionally accepts a list of slack channels
registered with AWS Chatbot. These channels receive notification upon
any failure in the pipeline.

-----


RomainMuller (approved):

lib/pipeline.ts
```diff
@@ -160,7 +161,13 @@ export interface PipelineProps {
...
    * @default - Duration.hours(8)
    */
-  buildTimeout?: Duration;
+  readonly buildTimeout?: Duration;
+
+  /**
+   * Slack channel configurations where failure notification
+   * in this pipeline should be sent.
+   */
+  readonly failureNotifySlack?: chatbot.SlackChannelConfiguration[];
```
rix0rrr, line 170: Would have preferred for failure notifications to have been an enum-like type, to pick between Slack and Chime .

lib/pipeline.ts
```diff
@@ -286,6 +293,23 @@ export class Pipeline extends Construct {
     if (props.autoBuild) {
       this.autoBuild(props.autoBuildOptions);
     }
+
+    if (props.failureNotifySlack && props.failureNotifySlack.length > 0) {
+      props.failureNotifySlack.forEach(s => {
+        new starnotifs.CfnNotificationRule(this, `FailureSlackNotification-${s.slackChannelConfigurationName}`, {
```
rix0rrr, line 299: Thanks CodeStar! Good find!

lib/pipeline.ts
```diff
@@ -286,6 +293,23 @@ export class Pipeline extends Construct {
...
+          name: `${this.pipeline.pipelineName}-failednotifications`,
+          detailType: 'BASIC',
```
rix0rrr, line 301: Now I'm curious about more options here.

lib/pipeline.ts
```diff
@@ -286,6 +293,23 @@ export class Pipeline extends Construct {
...
+          resource: this.pipeline.pipelineArn,
+          targets: [
+            {
```
rix0rrr, line 304: Wouldn't it have made more sense to put the array here instead of `forEach`ing over it? 

Or is this a fake array?

PR #551, opened 2020-11-05
chore(pipeline): refactor failure notification to use bind pattern

nija-at:
The pipeline notifications is now switched to use a bind pattern
with an enum-like class for better APIs.

This makes the API extensible with the ability for users to write
their own if they wish. It also makes the interface cleaner and
reduces property fragmentation on `PipelineProps` for the same
feature, i.e., failure notifications.

BREAKING CHANGES: The property `failureNotifySlack` has been
removed and replaced by `notifyOnFailure()` API.

-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -712,7 +712,7 @@ pipeline.notifyOnFailure(PipelineNotification.slack({
 
 // Chime
 const teamRoomWebhook = 'https://hooks.chime.aws/incomingwebhooks/1c3588c7-623d-4799-af9b-8b1818fca779?token=cUMzOVA4OXl8MXxCaHJlZ0RUVm03TmZVMkpoTzlwa3NVbXJCam8tNWF3UGdzemVqZndsZERV';
-pipeline.notifyOnFailure(PipelineNotification.slack({
+pipeline.notifyOnFailure(PipelineNotification.chime({
   webhookUrl: [ teamRoomWebhook ]
 }));
 ```
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ available:
    - [PyPi](#pypi)
    - [GitHub Releases](#github-releases)
    - [GitHub Pages](#github-pages)
+5. [Automatic Bumps and Pull Request Builds](#automatic-bumps-and-pull-request-builds)
+6. [Failure Notifications](#failure-notifications)
 
 
 ## Installation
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -694,6 +696,27 @@ NOTE: there is currently no way for the bump command to indicate to the
...
+const teamChannel = new chatbot.SlackChannelConfiguration(this, {
+  // ...
+});
+pipeline.notifyOnFailure(PipelineNotification.slack({
+  channels: [teamChannel]
+}));
+
+// Chime
+const teamRoomWebhook = 'https://hooks.chime.aws/incomingwebhooks/1c3588c7-623d-4799-af9b-8b1818fca779?token=cUMzOVA4OXl8MXxCaHJlZ0RUVm03TmZVMkpoTzlwa3NVbXJCam8tNWF3UGdzemVqZndsZERV';
+pipeline.notifyOnFailure(PipelineNotification.slack({
```
rix0rrr, line 715: Surely not `.slack` here?

lib/pipeline-notifications/chime.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Notify events on pipeline to a Chime room.
+ */
+export class ChimeNotification implements IPipelineNotification {
+  constructor(private readonly props: ChimeNotificationProps) {
+  }
+
+  public bind(options: PipelineNotificationBindOptions): void {
+    new ChimeNotifier(options.pipeline, 'PipelineNotificationChime', {
```
rix0rrr, line 17: This way we can only have 1. Is that intentional?

lib/pipeline-notifications/slack.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public bind(options: PipelineNotificationBindOptions): void {
+    const targets: starnotifs.CfnNotificationRule.TargetProperty[] = this.props.channels.map(c => {
+      return {
+        targetAddress: c.slackChannelConfigurationArn,
+        targetType: 'AWSChatbotSlack',
+      };
+    });
+    new starnotifs.CfnNotificationRule(options.pipeline, 'PipelineNotificationSlack', {
```
rix0rrr, line 34: Same here about only having 1.

lib/pipeline.ts
```diff
@@ -162,12 +160,15 @@ export interface PipelineProps {
...
-   * in this pipeline should be sent.
-   */
-  readonly failureNotifySlack?: chatbot.SlackChannelConfiguration[];
+export interface PipelineNotificationBindOptions {
+  readonly pipeline: Pipeline;
+  readonly codePipeline: cpipeline.Pipeline;
+}
+
+export interface IPipelineNotification {
+  bind(pipeline: PipelineNotificationBindOptions): void;
```
rix0rrr, line 171: Traditionally, `bind()` takes a `scope: IConstruct`, which is where it will create constructs.

Yes we'll call it with `bind(this, this)` but that still feels more correct and allows for more flexibility.

lib/pipeline.ts
```diff
@@ -293,23 +294,13 @@ export class Pipeline extends Construct {
...
-              targetType: 'AWSChatbotSlack',
-            },
-          ],
-          eventTypeIds: ['codepipeline-pipeline-action-execution-failed'],
-        });
-      });
-    }
+  public notifyOnFailure(notification: IPipelineNotification) {
+    notification.bind({
+      codePipeline: this.pipeline,
```
rix0rrr, line 301: Is `this.pipeline` not publicly accessible? If not, it probably ought to be.

After that, it sufficess to pass `this`.

rix0rrr (dismissed):
I want to do provisional approval but I don't think we can on this repo, so I'm turning it into Request Changes.

lib/pipeline-notifications/chime.ts
nija-at: Not intended. Is that because of the SingletonLambda?

rix0rrr (approved):

PR #554, opened 2020-11-06
chore: make Nuget and PyPI publishing respect the FOR_REAL variable

njlynch:
As part of investigating how to test publishing of V2, I discovered that neither
the Nuget or PyPI publishing scripts appear to handle the FOR_REAL env variable
that is respected by the other publishing steps. Copying the patterns from the
other publishing steps so these steps can be tested/mocked as well.

Related changes:
* I had to update the CodeBuild images, as the existing ones were deprecated and couldn't be used to launch new projects.

-----

### DIFF
--- a/lib/publishing/nuget/publish.sh
+++ b/lib/publishing/nuget/publish.sh
@@ -4,6 +4,17 @@ set -eu # we don't want "pipefail" to implement idempotency
 echo "Installing required CLI tools: jq, openssl..."
 yum install -y jq openssl
 
+if [[ "${FOR_REAL:-}" == "true" ]]; then
+    dotnet=dotnet
+else
+    echo "==========================================="
+    echo "            🏜️ DRY-RUN MODE 🏜️"
+    echo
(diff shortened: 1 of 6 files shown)


lib/publishing/nuget/publish.sh
```diff
@@ -4,6 +4,17 @@ set -eu # we don't want "pipefail" to implement idempotency
 echo "Installing required CLI tools: jq, openssl..."
 yum install -y jq openssl
 
+if [[ "${FOR_REAL:-}" == "true" ]]; then
+    dotnet=dotnet
```
rix0rrr, line 8: Doesn't this require the use of `$dotnet` somewhere down the line?

rix0rrr (changes requested):

lib/publishing/nuget/publish.sh
njlynch: D'oh! Of course it does. Fixed.

rix0rrr (approved):

PR #583, opened 2020-11-19
chore: fail NuGet publish step when there are errors

shivlaks:
the dotnet nuget push command gets the symbol packages, creates a PUT
request and then a response. Our pipeline does not currently fail on this command.

added `-o` flag to the `pipefail`. it was previously excluded for idempotency, which
is now being handled by the `--skip-duplicate` flag on the push.

Also added the `--force-english-output` flag which forces the application to run
using an invariant, English-based culture.

docs: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-push

-----


lib/publishing/nuget/publish.sh
```diff
@@ -86,6 +86,13 @@ for NUGET_PACKAGE_PATH in $(find dotnet -name *.nupkg -not -iname *.symbols.nupk
             # The .snupkg will be published at the same time as the .nupkg if both are in the current folder (which is the case)
             $dotnet nuget push $NUGET_PACKAGE_NAME -k $NUGET_API_KEY -s $NUGET_SOURCE | tee ${log}
         fi
+
+        # dotnet nuget push writes a stream of output. errors are prefixed with error:
+        # Alternatively we could check for success string "Your package was pushed."
+        if tail -n 1 ${log} | grep -q "error"; then
```
RomainMuller, line 92: This script runs *without* `-o pipefail`, and the `$dotnet nuget push` command above is piped to `tee ${log}`, meaning if the `$dotnet nuget push` command fails, it's exist status will be lost (in favor of `tee`'s, which is likely to always succeed).

Instead of doing this, I would suggest using the `${PIPESTATUS[n]}` array to check the exist status of the `$dotnet nuget push` command.

Now, we don't enable `-o pipefail` in order to preserve idempotency, and as-is, you'd also get an error if attempting to publish a package that is already there (not good!). In such cases, `nuget push` offers the `-SkipDuplicate` option that'll allow you to simply skip that package and move on to the next one to push... I reckon this is all we need (and hence we ought to be able to enable `-o pipefail` at the top of the script instead of doing this dance...

RomainMuller (changes requested):
That feels a little brittle to me... I would at least add `-ForceEnglishOutput` to the `nuget push` invocation.

But actually, there is a couple of better ways to do this... See the comment in the code.

lib/publishing/nuget/publish.sh
shivlaks: thanks @RomainMuller - that does sound a lot more robust. appreciate the guidance. 
updated - let me know if this is what you had in mind!

lib/publishing/nuget/publish.sh
```diff
@@ -80,11 +80,17 @@ for NUGET_PACKAGE_PATH in $(find dotnet -name *.nupkg -not -iname *.symbols.nupk
...
             # The .snupkg will be published at the same time as the .nupkg if both are in the current folder (which is the case)
-            $dotnet nuget push $NUGET_PACKAGE_NAME -k $NUGET_API_KEY -s $NUGET_SOURCE | tee ${log}
+            dotnet nuget push $NUGET_PACKAGE_NAME -k $NUGET_API_KEY -s $NUGET_SOURCE --force-english-output --skip-duplicate | tee ${log}
+        fi
+        
+        # we need to check PIPESTATUS since "tee" is the last command which is always likely to succeed
+        if [ ${PIPESTATUS[0]} -ne 0 ]; then
+            echo "❌ Error publishing ${NUGET_PACKAGE_NAME} to NuGet"
+            exit 1
         fi
```
RomainMuller, line 88: Since this now runs with `pipefail`, we need not check this anymore - the option's purpose is to make the whole pipe sequence result in an error if any of its stages is not exiting with status 0!

RomainMuller (approved):

lib/publishing/nuget/publish.sh
shivlaks: ahh i get it now

PR #587, opened 2020-11-20
feat: synchronize docker images to a local ECR repository

nija-at:
Refactor changes out of `aws/cdk-ops` so that it can be re-used.

Introduce an `EcrMirrorAspect` that can walk CodeBuild projects
and replace all uses of the dockerhub instance with the equivalents
in the ECR Mirror construct.

Motivation
EcrRegistrySync is also needed for `aws-delivlib` PR builds and
CodePipeline CI since they also use `jsii/superchain` as their base
image.

-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -750,7 +750,7 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
 
 ## ECR Mirror
 
-Builds commonly use Docker images, and these typically come from DockerHub. In fact, delivlib defaults its build
+Builds commonly use Docker images from DockerHub as their base image. In fact, delivlib defaults its build
 image to `jsii/superchain`. However, DockerHub has throttles in place for the volume of unauthenticated and
 authenticated pulls. This can cause CodeBuild jobs that run frequently to fail from DockerHub's throttling.
 
@@ -759,11 +759,11 @@ a local ECR registry in the AWS account.
 
 ```ts
 new EcrMirror(this, 'RegistrySync', {
-  images: [
+  sources: [
     MirrorSource.fromDockerHub('jsii/superchain'),
     MirrorSource.fromDockerHub('python:3.6'),
   ],
-  dockerhubCreds: // ...
+  dockerhubCredentials: // ...
   schedule: events.Schedule.cron( ... ),
 })
 ```
--- a/README.md
+++ b/README.md
@@ -758,11 +758,11 @@ The `EcrMirror` construct can be used to synchronize, on a specific schedule, Do
 a local ECR registry in the AWS account.
 
 ```ts
-new EcrRegistrySync(this, 'RegistrySync', {
-  images: MirrorSource.fromDockerHub([
-    'python:3.6',
-    'jsii/superchain'
-  ]),
+new EcrMirror(this, 'RegistrySync', {
+  images: [
+    MirrorSource.fromDockerHub('jsii/superchain'),
+    MirrorSource.fromDockerHub('python:3.6'),
+  ],
   dockerhubCreds: // ...
   schedule: events.Schedule.cron( ... ),
 })
--- a/README.md
+++ b/README.md
@@ -753,14 +753,11 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
 Builds commonly use Docker images, and these typically come from DockerHub. However, DockerHub has recently
 introduced throttles on their pulls. This causes CodeBuild jobs on high throughput repositories to be throttled.
 
-The `EcrRegistrySync` construct can be used to synchronize Docker images between DockerHub and a private ECR
+The `EcrRegistrySync` construct can be used to synchronize Docker images between DockerHub and the local ECR
 registry in the AWS account.
 
 ```ts
-const registry = `${cdk.Aws.ACCOUNT_ID}.dkr.ecr.${cdk.Aws.REGION}.amazonaws.com`;
-
 new EcrRegistrySync(this, 'RegistrySync', {
-  ecrRegistry: registry,
   images: ImageSource.fromDockerHub([
     'python:3.6',
     'jsii/superchain'
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ available:
    - [GitHub Pages](#github-pages)
 1. [Automatic Bumps and Pull Request Builds](#automatic-bumps-and-pull-request-builds)
 1. [Failure Notifications](#failure-notifications)
-1. [ECR Registry Sync](#ecr-registry-sync)
+1. [ECR Mirror](#ecr-mirror)
 
 
 ## Installation
@@ -748,13 +748,14 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
 }));
 ```
 
-## ECR Registry Sync
+## ECR Mirror
 
-Builds commonly use Docker images, and these typically come from DockerHub. However, DockerHub has recently
-introduced throttles on their pulls. This causes CodeBuild jobs on high throughput repositories to be throttled.
+Builds commonly use Docker images, and these typically come from DockerHub. In fact, delivlib defaults its build
+image to `jsii/superchain`. However, DockerHub has throttles in place for the volume of unauthenticated and
+authenticated pulls. This can cause CodeBuild jobs that run frequently to fail from DockerHub's throttling.
 
-The `EcrRegistrySync` construct can be used to synchronize Docker images between DockerHub and the local ECR
-registry in the AWS account.
+The `EcrMirror` construct can be used to synchronize, on a specific schedule, Docker images between DockerHub and
+a local ECR registry in the AWS account.
 
 ```ts
 new EcrRegistrySync(this, 'RegistrySync', {
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ available:
    - [GitHub Pages](#github-pages)
 1. [Automatic Bumps and Pull Request Builds](#automatic-bumps-and-pull-request-builds)
 1. [Failure Notifications](#failure-notifications)
+1. [ECR Registry Sync](#ecr-registry-sync)
 
 
 ## Installation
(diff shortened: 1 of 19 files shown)


iliapolo (dismissed):

README.md
```diff
@@ -748,13 +748,14 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
 }));
 ```
 
-## ECR Registry Sync
+## ECR Mirror
```
iliapolo, line 751: I've been struggling with this name originally as well. Its not really a mirror because it can build custom images. Also mirror gives the impressions its fully mirroring an upstream, which it isn't.

I landed on `EcrRegistry` since it basically represents a registry that users can push images to. The fact some of them are coming from `GitHub` is somewhat incidental now and is manifested in the `ImageSource` class, which i feel is the right place.

The fact it can periodically update the images is also just a registry feature, not necessarily related to any upstream.

WDTYT?

lib/registry-sync/ecr-mirror.ts
```diff
@@ -187,14 +185,8 @@ export class EcrRegistrySync extends Construct {
 
   }
 
-  /**
-   * Grant the specified grantees pull privileges to the target ECR repositories.
-   */
-  public grantPull(...grantees: iam.IGrantable[]) {
```
iliapolo: Can you explain the reasoning behind removing this? This was rather useful from what I recall.

lib/registry-sync/ecr-mirror.ts
```diff
@@ -187,14 +185,8 @@ export class EcrRegistrySync extends Construct {
...
-    for (const grantee of grantees) {
-      this._grantAuthorize(grantee);
-      this._repos.forEach(p => p.grantPull(grantee));
-    }
+  public ecrRepository(source: MirrorSource): ecr.IRepository | undefined {
+    return this._repos.get(source);
```
iliapolo, line 189: Doesn't this mean:

```ts

const superchain1 = MirrorSource.fromDockerHub('jsii/superchain');
const superchain2 = MirrorSource.fromDockerHub('jsii/superchain');

const mirror = new EcrMirror(this, 'Mirror', {
  ...,
  images: [superchain1]
})

const repo1 = mirror.ecrRepository(superchain1); // works
const repo2 = mirror.ecrRepository(superchain2); // fails 
```

This is unexpected behavior IMO and can lead to some quirky bugs.

I think is better to implement a `toString()` on the `MirrorSource` class and have `this._repos` be a `Map<string, ecr.IRepository>`.

pipeline/delivlib.ts
```diff
@@ -81,11 +78,66 @@ export class DelivLibPipelineStack extends Stack {
...
+      if (!Token.isUnresolved(cfnproject.environment)) {
+        const env = cfnproject.environment as codebuild.CfnProject.EnvironmentProperty;
+        if (env.image === 'jsii/superchain' || env.image === 'jsii/superchain:latest') {
+          cfnproject.environment = {
+            ...env,
+            image: codebuild.LinuxBuildImage.fromEcrRepository(this.superchainRepo).imageId,
+          };
+          this.superchainRepo.grantPull(construct);
+          construct.grantPrincipal.addToPrincipalPolicy(new iam.PolicyStatement({
+            actions: ['ecr:GetAuthorizationToken'],
```
iliapolo, line 121: This knowledge feels more appropriate in the mirror itself. Perhaps the mirror should offer `mirror.grantPullRepository` ? or `mirror.grantPullImage`.

pipeline/delivlib.ts
```diff
@@ -81,11 +78,66 @@ export class DelivLibPipelineStack extends Stack {
...
+      throw new Error('Cannot find ECR mirror repository for "jsii/superchain"');
+    }
+    this.superchainRepo = repo;
+  }
+}
+
+export class EcrMirrorAspect implements IAspect {
+  constructor(private readonly superchainRepo: ecr.IRepository) {}
+
+  public visit(construct: IConstruct) {
```
iliapolo, line 109: This is cool, from a technical perspective, but actually exposes a problem with non configurable build images in delivlib. Is it a big effort to make them configurable and just pass the ECR image?

lib/registry-sync/ecr-mirror.ts
nija-at: The implementation `grantPull()` grants permissions to all ECR repos, which doesn't follow the 'least privileges' policy.
With the `ecrRepository()` API, it can be easily done by getting the corresponding repo and calling `grantPull` on it.

nija-at: I'm wondering if we should elevate `grantAuthorize()` to the `aws-ecr` module?

nija-at: https://github.com/aws/aws-cdk/pull/11783

README.md
nija-at: I would avoid over-thinking the name and trying to over-generalize the set of features. What we're aiming to do here is to sync docker images into ECR.

How about removing the `fromDirectory()` API from delivlib since it's only needed for `cdk-ops`, and then the construct will more closely match an ECR Mirror construct.

lib/registry-sync/ecr-mirror.ts
nija-at: Fair concern. I have the same one. I'll improve.

pipeline/delivlib.ts
nija-at: Why should the `grantPull()` go into the mirror construct? It's correctly placed as an API of ECR `Repository` construct.

pipeline/delivlib.ts
nija-at: There's too many places this occurs - autobuild, autoPR, mergeback, regular builds. Trying to pull them all out from underneath the layers isn't trivial.

Isn't this a good use case for an aspect? I'm even wondering if we should export this out of delivlib.

README.md
iliapolo: I think removing `fromDirectory` will be a little strange. The existence of the `MirrorSource` class and the `bind` API you've added makes it inherently extensible, which is great, but it is no longer just a mirror. 

> What we're aiming to do here is to sync docker images into ECR

This was the initial intent, but it grew because we needed it to, and it then ended up as a generic construct that allows pushing images to an ECR registry. 

In any case, I don't mind keeping the name, just wanted to offer my point of view.

lib/registry-sync/ecr-mirror.ts
iliapolo: > aws/aws-cdk#11783

👍 

> The implementation grantPull() grants permissions to all ECR repos, which doesn't follow the 'least privileges' policy.

Ok, I think it's probably ok though to treat the registry as a single grantable unit, especially since its only read, but lets see if this becomes tricky in case the caller doesn't know which repos it needs (granted - not the ideal scenario).

pipeline/delivlib.ts
iliapolo: To hide the IAM knowledge from the caller, but I agree aws/aws-cdk#11783 is better.

pipeline/delivlib.ts
iliapolo: I had always considered Aspects to be more of an augmentation/validation, rather than mutation. Personally I find this a little unexpected.

But thats fine, we can get back to it at some point.

lib/registry-sync/ecr-mirror.ts
```diff
@@ -0,0 +1,223 @@
...
+  readonly autoStart?: boolean;
+}
+
+/**
+ * Synchronize images from DockerHub to an ECR registry in the AWS account.
+ * This is particularly useful to workaround DockerHub's throttling on pulls and use ECR instead.
+ */
+export class EcrMirror extends Construct {
+
+  private readonly _repos: Map<string, ecr.IRepository> = new Map();
```
iliapolo, line 77:
```suggestion
  private readonly _repos: Map<string, ecr.Repository> = new Map();
```

No need to potentially limit the API available to consumers, this will always be an owned repository.

README.md
```diff
@@ -747,6 +748,39 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
 }));
 ```
 
+## ECR Mirror
+
+Builds commonly use Docker images, and these typically come from DockerHub. In fact, delivlib defaults its build
```
RomainMuller, line 753:
```suggestion
Builds commonly used Docker images, and these typically come from DockerHub. In fact, delivlib defaults its build
```

README.md
```diff
@@ -747,6 +748,39 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
...
+
+The `EcrMirror` construct can be used to synchronize, on a specific schedule, Docker images between DockerHub and
+a local ECR registry in the AWS account.
+
+```ts
+new EcrMirror(this, 'RegistrySync', {
+  images: [
+    MirrorSource.fromDockerHub('jsii/superchain'),
+    MirrorSource.fromDockerHub('python:3.6'),
+  ],
```
RomainMuller, line 765: Optional: since the type of content is `MirrorSource`, then should this key be called `sources:`?

README.md
```diff
@@ -747,6 +748,39 @@ pipeline.notifyOnFailure(PipelineNotification.chime({
...
+  dockerhubCreds: // ...
```
RomainMuller, line 766: Can we avoid abbreviating?

lib/registry-sync/ecr-mirror.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Sync job runs on a schedule.
+   * @default - does not run on schedule
+   */
+  readonly schedule?: events.Schedule;
+
+  /**
+   * Start the sync job immediately after the deployment.
+   * @default false
```
RomainMuller, line 65: I reckon this involves injecting a CustomResource? Maybe mention that?

lib/registry-sync/ecr-mirror.ts
```diff
@@ -0,0 +1,222 @@
...
+ */
+export class EcrMirror extends Construct {
+
+  private readonly _repos: Map<string, ecr.Repository> = new Map();
+  private readonly _project: codebuild.Project;
+
+  constructor(scope: Construct, id: string, props: EcrMirrorProps) {
+    super(scope, id);
+
+    const ecrRegistry = `${Stack.of(scope).account}.dkr.ecr.${Stack.of(scope).region}.amazonaws.com`;
```
RomainMuller, line 82: Would be noice if this could be factored into the `@aws-cdk/aws-ecr.IRepository` contract... Because this is duplicated in so many places...

lib/registry-sync/ecr-mirror.ts
```diff
@@ -0,0 +1,222 @@
...
+    const commands = [];
```
RomainMuller, line 83: Does this not get type-inferred as a `never[]`? Anyhow this looks inconsistent with the line right below.

test/registry-sync/docker-asset/Dockerfile
```diff
@@ -0,0 +1 @@
+# empty for now
```
RomainMuller, line 1: For now?

RomainMuller (approved):
Nothing blocking, so granting approval, but I'll mark as `pr/do-not-merge` as I have suggestions...

README.md
nija-at: It should be a statement in the present, and not in the past.

I'm going to adjust it this way instead.

"Builds commonly use Docker images from DockerHub as their base image."

README.md
nija-at: Good call.

lib/registry-sync/ecr-mirror.ts
nija-at: Don't want to wait for another release of the CDK here. Already did something similar here - https://github.com/aws/aws-cdk/pull/11783

I've opened an issue to track this for now https://github.com/aws/aws-cdk/issues/11915

lib/registry-sync/ecr-mirror.ts
nija-at: it gets inferred as `any[]`. Fixed to be `string[]`.

PR #633, opened 2020-12-17
feat(auto-build): option to not publish build logs when build is successful

jogold:
Removes noise in PR comments.

I bet those logs are hardly ever checked.

See https://github.com/jlhood/github-codebuild-logs/pull/29

-----

### DIFF
--- a/lib/auto-build.ts
+++ b/lib/auto-build.ts
@@ -34,7 +34,7 @@ export interface AutoBuildOptions {
    *
    * @see https://github.com/jlhood/github-codebuild-logs#app-parameters
    *
-   * @default false
+   * @default true
    */
   readonly publicLogsOnSuccess?: boolean;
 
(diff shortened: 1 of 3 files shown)


lib/auto-build.ts
```diff
@@ -29,6 +29,15 @@ export interface AutoBuildOptions {
...
   readonly publicLogs?: boolean;
 
+  /**
+   * Whether to publish a link to build logs when build is successful.
+   *
+   * @see https://github.com/jlhood/github-codebuild-logs#app-parameters
+   *
+   * @default false
+   */
+  readonly publicLogsOnSuccess?: boolean;
```
eladb, line 39: Let's make the default true. I find it very useful to get an email when my build passes.

eladb (changes requested):

lib/auto-build.ts
jogold: 😄 this is exactly the kind of noise I was referring to in the PR description 😄 

I personally don't like to get 5 emails when I push 5 different commits to a PR... and I like to only get alerts/notifications when something fails (this is the behavior of failed GitHub workflows).

Tell me if you still want to change this default, I will of course adapt accordingly 😉

lib/auto-build.ts
```diff
@@ -90,11 +99,12 @@ export class AutoBuild extends Construct {
       new serverless.CfnApplication(this, 'GitHubCodeBuildLogsSAR', {
         location: {
           applicationId: 'arn:aws:serverlessrepo:us-east-1:277187709615:applications/github-codebuild-logs',
-          semanticVersion: '1.3.0',
+          semanticVersion: '1.4.0',
```
NetaNir, line 102: Oh we really need this upgrade... (we really need this to be configurable, and we need code build to do be better but I'll take what I can get)

lib/auto-build.ts
eladb: Let's change the default to true

eladb (approved):

PR #671, opened 2021-01-11
chore(pull-request): remove the use of Omit chain

nija-at:
We've introduced an Omit chain in the interfaces around auto pull 
requests. Instead, use more traditional 'Props inherits Options'
approach to make the code more readable.

----

#### Additional Testing

Applied this with change on `cdk-ops` using `yarn link` and noted that there were no visible changes that seemed to indicate a diff here. 

-----

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -18,7 +18,8 @@ import { ChangeController } from './change-controller';
 import { ChimeNotifier } from './chime-notifier';
 import { PipelineWatcher } from './pipeline-watcher';
 import * as publishing from './publishing';
-import { AutoBump, AutoMergeBack, AutoMergeBackProps, AutoBumpProps } from './pull-request';
+import { AutoBump, AutoMergeBack, AutoBumpProps } from './pull-request';
+import { AutoMergeBackOptions } from './pull-request/merge-back';
 import { IRepo, WritableGitHubRepo } from './repo';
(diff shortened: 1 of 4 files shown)


iliapolo:
@nija-at So as I understand it, you basically replaced:

```ts
Omit<AutoPullRequestProps, 'commands' | 'condition' | 'head'>
``` 

with a concrete `AutoPullRequestOptions` interface and made `AutoPullRequestProps` extend `AutoPullRequestOptions`.

Right? 

But what if we want to add another specific type of PR that actually doesn't want to override `commands`? The nice thing about `Omit` is that it doesn't force any premature generalization. I'm not sure `AutoPullRequestOptions` will be applicable to all types of future PR's we might want to add.

Having said that, I don't mind merging this because it does fit the current use case and I agree its more readable, But just wondering if you considered that.

nija-at:
> But what if we want to add another specific type of PR that actually doesn't want to override commands?

In general, I would find 1 or 2 common base classes that most PR 'types' would want to use. 
In your example, we could create a `AutoPullRequestCommandOptions` that inherits `AutoPullRequestOptions` and has the `command` property included.
Beyond that there will be some duplication where the same property is repeated in two different props structures.

The latter feels better than an omit chain, where it gets quite hard to interpret what properties are part of which structures.
Added bonus: delivlib is closer to being jsii compat.

iliapolo (approved):

PR #684, opened 2021-01-18
fix: remove repo as required property of AutoMergeBackOptions

njlynch:
The refactoring in #671 unintentionally changed the signature of
AutoMergeBackOptions to make `repo` a required property, and also
added the `stage` property to AutoMergeBackProperties. This change
reverts both back to the pre-refactor state.

-----

### DIFF
--- a/lib/pull-request/merge-back.ts
+++ b/lib/pull-request/merge-back.ts
@@ -39,9 +39,8 @@ export interface MergeBackStage {
 
 /**
  * AutoMergeBackOptions and AutoMergeBackProps both share all of these options,
- * but AutoMergeBackProps can't extend AutoMergeBackOptions, as the 'stage' prop of AutoMergeBackOptions
- * is only applicable to the AutoMergeBackOptions interface. This interface instead captures all of the
- * common props between the two.
+ * but AutoMergeBackProps can't extend AutoMergeBackOptions; the latter has a 'stage' prop the former doesn't have.
+ * This interface instead captures all of the common props between the two.
  */
 interface CommonMergeOptions extends pr.AutoPullRequestOptions {
   /**
@@ -89,7 +88,7 @@ interface CommonMergeOptions extends pr.AutoPullRequestOptions {
   condition?: string;
 }
 
-export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions, CommonMergeOptions {
+export interface AutoMergeBackOptions extends CommonMergeOptions {
   /**
    * Specify stage options to create the merge back inside a stage of the pipeline.
    *
@@ -98,7 +97,7 @@ export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions, CommonM
   readonly stage?: MergeBackStage
 }
 
-export interface AutoMergeBackProps extends pr.AutoPullRequestOptions, CommonMergeOptions {
+export interface AutoMergeBackProps extends CommonMergeOptions {
   /**
    * The repository to bump.
    */
--- a/lib/pull-request/merge-back.ts
+++ b/lib/pull-request/merge-back.ts
@@ -37,7 +37,13 @@ export interface MergeBackStage {
   readonly after: string;
 }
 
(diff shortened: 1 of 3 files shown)


lib/pull-request/merge-back.ts
```diff
@@ -81,7 +87,9 @@ export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions {
    * @default - no condition
    */
   condition?: string;
+}
 
+export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions, CommonMergeOptions {
```
nija-at, line 92: I believe multiple inheritance is not jsii supported. While delivlib is not jsii enabled, might be nice to try to strive for that modeling.

nija-at: On second look - it looks like `CommonMergeOptions` already extends `AutoPullRequestOptions` so probably just drop the redundant one.

lib/pull-request/merge-back.ts
```diff
@@ -37,7 +37,13 @@ export interface MergeBackStage {
   readonly after: string;
 }
 
-export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions {
+/**
+ * AutoMergeBackOptions and AutoMergeBackProps both share all of these options,
+ * but AutoMergeBackProps can't extend AutoMergeBackOptions, as the 'stage' prop of AutoMergeBackOptions
+ * is only applicable to the AutoMergeBackOptions interface. This interface instead captures all of the
```
nija-at, line 43: > as the 'stage' prop of AutoMergeBackOptions is only applicable to the AutoMergeBackOptions interface

I don't understand this statement. Where is the `stage` property needed and not needed?

lib/pull-request/merge-back.ts
njlynch: Yeah, I'll drop the redundant `AutoPullRequestOptions` inheritence; that was a mistake.

lib/pull-request/merge-back.ts
njlynch: Reworded.

lib/pull-request/merge-back.ts
```diff
@@ -37,7 +37,12 @@ export interface MergeBackStage {
   readonly after: string;
 }
 
-export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions {
+/**
+ * AutoMergeBackOptions and AutoMergeBackProps both share all of these options,
+ * but AutoMergeBackProps can't extend AutoMergeBackOptions; the latter has a 'stage' prop the former doesn't have.
+ * This interface instead captures all of the common props between the two.
+ */
+interface CommonMergeOptions extends pr.AutoPullRequestOptions {
```
nija-at, line 45: How about something like this?

AutoMergeBackProps will extend AutoMergeBackOptions, neither of which will have stage.
Another struct, say StageMergeBackOption extends AutoMergeBackOptions and includes stage.

A few things become cleaner this way - this CommomMergeBackOptions goes away which sounds awkward. And AutoMergeBackProps extends AutoMergeBackOptions which is more natural in how we model construct props.

Will that work?

njlynch: I like this -- and considered something similar -- except it removes `stage` from the `AutoMergeBackOptions` interface, which would be backwards-incompatible. Now practically, as a JS/TS user of this library, it would largely be invisible (as long as the pipelines interface is changed to a `StageMergeBackOptions`), but it does seem like it would require a major version bump to be semver-compliant. Meanwhile, the point of this change was to fix a backwards-incompatible change which broke semver for our consumers, so doing another backwards-incompatible change here feels wrong.

(After discussing offline) - Let's ship this as-is, then do your suggestion as part of a major version bump.

nija-at:
Discussed offline that this will fix the bug and keep the interfaces backwards compatible.

As a follow-up, the correct interfaces will be included and released as a new major version

nija-at (approved):

PR #687, opened 2021-01-19
chore(bump): refactor AutoMergeBackOptions

njlynch:
As discussed in #684, refactoring the AutoMergeBack interfaces to have a more
standard hierarchy. This change removes 'stage' from AutoMergeBackOptions,
and creates a stage-specific AutoMergeBackPipelineOptions.

BREAKING CHANGE: the 'stage' prop of AutoMergeBackOptions has been moved to a
new AutoMergeBackPipelineOptions interface.

-----

### DIFF
--- a/lib/pull-request/merge-back.ts
+++ b/lib/pull-request/merge-back.ts
@@ -83,7 +83,7 @@ export interface AutoMergeBackOptions extends pr.AutoPullRequestOptions {
   condition?: string;
 }
 
-export interface AutoMergeBackOptionsWithStage extends AutoMergeBackOptions {
+export interface AutoMergeBackPipelineOptions extends AutoMergeBackOptions {
   /**
    * Specify stage options to create the merge back inside a stage of the pipeline.
    *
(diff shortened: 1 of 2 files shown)


lib/pull-request/merge-back.ts
```diff
@@ -88,7 +83,7 @@ interface CommonMergeOptions extends pr.AutoPullRequestOptions {
   condition?: string;
 }
 
-export interface AutoMergeBackOptions extends CommonMergeOptions {
+export interface AutoMergeBackOptionsWithStage extends AutoMergeBackOptions {
```
nija-at, line 86: Would StageMergeBackOptions or AutoMergeBackStageOptions. I prefer the former.

njlynch: `StageMergeBackOptions` sounds like we're merging back from a stage. `AutoMergeBackStageOptions` could be ok, I guess. How about `AutoMergeBackPipelineOptions`, so if we add something else besides a stage in the future, we don't have to rename again?

nija-at (approved):

PR #696, opened 2021-01-27
feat(pipeline): improved pipeline failure metrics

nija-at:
Currently, metrics are emitted for each pipeline failure as
`<pipelineName>_FailedStages`. This metric is emitted per pipeline once
per minute depending on the state of the pipeline at the time.

This means that if a pipeline is in failed state for 20 minutes, there
will be 20 non-zero datapoints for this metric.
This is not very useful in aggregating failure over time and summarizing
them over hours, days or weeks.

This commit changes the approach entirely. Instead of running the
metrics on a fixed schedule, it reacts to CodePipeline events and emits
the appropriate metric. This should result in one non-zero data point
per failure.

This commit also changes the metric name and includes dimensions for
better querying. It now emits a single metric `PipelineActionFailures`
under the namespace `CDK/Delivlib`. This metric has two dimensions -
`Pipeline` with the name of the pipeline and `Action` with the name of
the Action.
This lets higher resolution on the metrics to identify specific Actions
that are failing more often, while letting aggregated metrics.

BREAKING CHANGE: delivlib `Pipeline` construct no longer produces
the `<pipelineName>_FailedStages` metric. It instead produces the
metric `PipelineActionFailures` with the pipeline name as a dimension.

-----

See https://github.com/awslabs/aws-delivlib/pull/696#issuecomment-770991135
for additional testing

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,11 +674,21 @@ To create an ssh deploy key for your repository:
 The `Pipeline` construct automatically creates the following metrics in CloudWatch
 for the configured pipelines. These are published under the namespace 'CDK/Delivlib'.
 
-- *PipelineActionFailures:* For every failed action in the pipeline, a '1' is
-  published to the metric, and for every successful action, sa '0' is published.
+- Execution Failures: The number of failures of the pipeline execution.
+  When a pipeline execution fails, a '1' is recorded and forevery success, a '0' is
+  recorded.
 
-  This metric has the following dimensions:
-  - *Pipeline*: The pipeline name in CodePipeline,
+  Metric Name: *Failures*
+  Dimensions:
+  - *Pipeline*: The pipeline name in CodePipeline.
+
+- Action Failures: The number of failures per action per pipeline. An execution
+  failure can be due to multiple actions failing.
+  For every action failure, a '1' is recorded and for every success, a '0' is recorded.
+
+  Metric Name: *Failures*
+  Dimensions:
+  - *Pipeline*: The pipeline name in CodePipeline.
   - *Action*: THe name of the action that succeeded or failed.
 
 ## Automatic Bumps and Pull Request Builds
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ available:
    - [PyPi](#pypi)
    - [GitHub Releases](#github-releases)
    - [GitHub Pages](#github-pages)
+1. [Metrics](#metrics)
 1. [Automatic Bumps and Pull Request Builds](#automatic-bumps-and-pull-request-builds)
 1. [Failure Notifications](#failure-notifications)
 1. [ECR Mirror](#ecr-mirror)
@@ -667,6 +668,18 @@ To create an ssh deploy key for your repository:
 1. Create an AWS Secrets Manager secret and paste the private key as plaintext
    (not key/value).
 1. Use the name of the AWS Secrets Manager secret in the `sshKeySecret` option.
+
+## Metrics
+
+The `Pipeline` construct automatically creates the following metrics in CloudWatch
+for the configured pipelines. These are published under the namespace 'CDK/Delivlib'.
+
+- *PipelineActionFailures:* For every failed action in the pipeline, a '1' is
+  published to the metric, and for every successful action, sa '0' is published.
+
+  This metric has the following dimensions:
+  - *Pipeline*: The pipeline name in CodePipeline,
+  - *Action*: THe name of the action that succeeded or failed.
 
 ## Automatic Bumps and Pull Request Builds
 
--- a/lib/pipeline-watcher/handler/watcher-handler.ts
+++ b/lib/pipeline-watcher/handler/watcher-handler.ts
@@ -15,11 +15,8 @@ const logger = {
 };
 
 /**
- * Lambda function for checking the stages of a CodePipeline and emitting log
- * entries with { failedCount = <no. of failed stages> } for async metric
- * aggregation via metric filters.
- *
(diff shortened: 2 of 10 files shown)


nija-at:
### Additional Testing Notes

Deployed these changes to an AWS account, running the [test stack](https://github.com/awslabs/aws-delivlib/blob/main/test/test-stack.ts) and verified the metrics produced.

Use account 712950704752 and see for yourself [here](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2:graph=~(metrics~(~(~(expression~'SEARCH*28*27*7bCDK*2fDelivlib*2cAction*2cPipeline*7d*20MetricName*3d*22Failures*22*27*2c*20*27Sum*27*2c*20300*29~id~'e1~label~'Action))~(~'CDK*2fDelivlib~'Failures~'Pipeline~'delivlib-test-CodeCommitPipelineBuildPipeline656B8CCB-1G936MI8U73K~(id~'m1~yAxis~'right~label~'Pipeline*20Failures~stat~'Sum)))~view~'timeSeries~stacked~false~region~'us-east-1~start~'2021-01-29T11*3a00*3a00.000Z~end~'2021-01-29T12*3a37*3a00.000Z~stat~'Average~period~300);query=~'*7bCDK*2fDelivlib*2cAction*2cPipeline*7d*20MetricName*3d*22Failures*22).

rix0rrr:
I am not sure I dig this. I prefer level-triggered metrics to edge-triggered metrics for alarming... although I have to reverse-engineer the reasons now, my spidey senses are tingling here.

You're saying the aggregate of `0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 1, 1` isn't useful... but I disagree. `MAX()` can definitely be made to be useful, for example.

Why not add these new metrics in addition to the old ones, to start with?

Also, are you sure `Dimensions` do what you think they do? Fairly sure they are just labeled additions to the key, and they don't auto-aggregate on arbitrary subsets of the dimensions. You must specify *exactly* the same Dimensions values to get the values out again. That is to say, if you emit metrics under `{ Pipeline, Action }`, you **cannot** query on just `{ Pipeline }` alone. 

To get the behavior you want, you probably have to emit 3 datapoints:

| Metric | Dimensions | Value |
|----|----|----|
| PipelineActionFailures | | 1 |
| PipelineActionFailures | Pipeline=$pipeline | 1 |
| PipelineActionFailures | Pipeline=$pipeline,Action=$action | 1 |

nija-at:
> Why not add these new metrics in addition to the old ones, to start with?

AFAICT, there's one alarm we're configuring based on level metrics. This can be just easily configured on counter metrics (as has been done in this PR). It seems that the alarm will behave exactly the same way (since it's told to ignore missing data points).

The only gap I can tell is when there's an error in the lambda and metrics fail to emit. Nothing will let us know of this. If this is a concern, we can cover that gap with another alarm that checks for metrics existence (I believe this is not any different today).

> To get the behavior you want, you probably have to emit 3 datapoints

Actually, just the last two are sufficient, (and those are what I'm emitting with this change). These should answer both pipeline level information, and action level information. 
The first one (in your table) is an aggregate of "all" pipeline failures - not sure how that's useful.

To your point about using `Dimensions` correctly - this is the first time I'm emitting CW metrics so I've had to read up and experiment to figure out the correct use. There's a chance I got it wrong.

The link in the 'Additional Testing Notes' section written above has test metrics I've emitted. Maybe that'll show you how the data will look when this is merged.

rix0rrr:
> It seems that the alarm will behave exactly the same way

If you're only logging `1`s, not logging `0`s, and the time between emissions is unpredictable, how will the alarm deactivate?

rix0rrr:
> those are what I'm emitting with this change

Oh I see. I expected those to be emitted together, but they're each emitted because of different subscriptions. Correct?

rix0rrr (approved):
I'm okay shipping this after you confirm the alarm will correct activate and deactivate based on the new metrics.

nija-at:
>> It seems that the alarm will behave exactly the same way
>
> If you're only logging 1s, not logging 0s, and the time between emissions is unpredictable, how will the alarm deactivate?

I'm depending on the behaviour of how missing data points are handled. From [the docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data), "ignore – The current alarm state is maintained "

Does that satisfy?

nija-at:
### Additional Testing Notes

Deployed these changes to an AWS account, running the [test stack](https://github.com/awslabs/aws-delivlib/blob/main/test/test-stack.ts)
and verified the metrics produced.

Use account 712950704752 and see for yourself [here](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2:graph=~(metrics~(~(~(expression~'SEARCH*28*27*7bCDK*2fDelivlib*2cAction*2cPipeline*7d*20MetricName*3d*22Failures*22*27*2c*20*27Sum*27*2c*20300*29~id~'e1~label~'Action))~(~'CDK*2fDelivlib~'Failures~'Pipeline~'delivlib-test-CodeCommitPipelineBuildPipeline656B8CCB-1G936MI8U73K~(id~'m1~yAxis~'right~label~'Pipeline*20Failures~stat~'Sum)))~view~'timeSeries~stacked~false~region~'us-east-1~start~'2021-01-29T11*3a00*3a00.000Z~end~'2021-01-29T12*3a37*3a00.000Z~stat~'Average~period~300);query=~'*7bCDK*2fDelivlib*2cAction*2cPipeline*7d*20MetricName*3d*22Failures*22).

PR #730, opened 2021-02-23
feat: golang publishing

eladb:
Add support for publishing golang code from the build artifact (under the `go/` directory) to a GitHub repository.

To add a go publishing step to your pipeline, use `pipeline.publishToGolang()` or create an instance of `PublishToGolang` construct.

See the README of [jsii-release] for more details.

**MISC**: allow `Shellable` environment values to be `undefined`, and then they will not be set.

**TESTING**: updated [awslabs/aws-delivlib-sample](https://github.com/awslabs/aws-delivlib-sample) to include a `go` section in the jsii config and tested that the delivlib test pipeline publishes go bindings to the same repo (under the [golang](https://github.com/awslabs/aws-delivlib-sample/tree/golang) branch).

Resolves aws/jsii#2562

[jsii-release]: https://github.com/aws/jsii-release

-----

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -398,6 +398,9 @@ export class Pipeline extends Construct {
     }), options);
   }
 
+  /**
+   * Publish Golang code from `go` directory in build artifact to a GitHub repository.
+   */
   public publishToGolang(options: publishing.PublishToGolangProps) {
     this.addPublish(new publishing.PublishToGolang(this, 'Golang', {
       dryRun: this.dryRun,
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -398,6 +398,13 @@ export class Pipeline extends Construct {
     }), options);
   }
 
+  public publishToGolang(options: publishing.PublishToGolangProps) {
+    this.addPublish(new publishing.PublishToGolang(this, 'Golang', {
+      dryRun: this.dryRun,
+      ...options,
+    }));
+  }
+
   /**
    * Enables automatic bumps for the source repo.
    * @param options Options for auto bump (see AutoBumpOptions for description of defaults)
--- a/lib/util.ts
+++ b/lib/util.ts
@@ -40,14 +40,16 @@ export function hashFileOrDirectory(fileOrDir: string): string {
   return hash.digest('base64');
 }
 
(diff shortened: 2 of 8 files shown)


lib/pipeline.ts
```diff
@@ -398,6 +398,13 @@ export class Pipeline extends Construct {
     }), options);
   }
 
+  public publishToGolang(options: publishing.PublishToGolangProps) {
```
RomainMuller, line 401: That name is... awkward. Other functions are named `publishTo<PackageRegistry>`, this one is `publishTo<Language>`. I guess it may be better off named `publishGoBindingsToGitRepository` or something...

Not exactly important, though.

lib/util.ts
```diff
@@ -40,14 +40,16 @@ export function hashFileOrDirectory(fileOrDir: string): string {
...
-export function renderEnvironmentVariables(env?: { [key: string]: string }, type?: cbuild.BuildEnvironmentVariableType) {
+export function renderEnvironmentVariables(env?: { [key: string]: string | undefined }, type?: cbuild.BuildEnvironmentVariableType) {
   if (!env) {
     return undefined;
   }
 
   const out: { [key: string]: cbuild.BuildEnvironmentVariable } = { };
   for (const [key, value] of Object.entries(env)) {
-    out[key] = { value, type };
+    if (value) {
```
RomainMuller, line 50: Maybe intentional - but are you certain you want to treat `""` as `undefined` here? Because both are false-y (this may effectively be the same thing though - not sure how CodeBuild behaves with empty string values).

lib/pipeline.ts
eladb: I'll leave like this. We use `golang` when it comes to publishing go bindings to github in jsii-release (and projen) so I wanted  to stay consistent.

lib/util.ts
eladb: Mmm.. good point. Let me fix.

eladb: Fixed

lib/pipeline.ts
eladb, line 401:
```suggestion
  /**
   * Publish Golang code from `go` directory in build artifact to a GitHub repository.
   */
  public publishToGolang(options: publishing.PublishToGolangProps) {
```

eladb: Added a docstring with some details.

RomainMuller (approved):

PR #739, opened 2021-03-05
feat: support NPM publishing order

rix0rrr:
Read NPM packages to be published from an `js/npm-publish-order.txt`
file, and publish packages in that order.

Tarballs found that are NOT in that list will be published afterwards.

-----


lib/publishing/npm/publish-npm.sh
```diff
@@ -40,8 +40,20 @@ if echo ${TOKENS} | grep 'EAUTHUNKNOWN' > /dev/null; then
...
+# Get a list of tarballs to publish, in order:
+# 1. Print the 'js/npm-publish-order.txt' file if it exists;
+# 2. List all tarballs in the js/ directory (strip leading './')
+# 3. Retain only unique lines from both of these (with awk magic: https://stackoverflow.com/a/11532197).
+#
+# Result is that the js/npm-publish-order.txt files are published in indicated
+# order, and all remaining tarballs not in that file afterwards.
+list_of_tarballs=$(cd ${PWD}/js && { \
+    [[ ! -f npm-publish-order.txt ]] || cat npm-publish-order.txt && \
+    find . -iname '*.tgz' | sed 's/^\.\///'; } | awk '!x[$0]++')
```
nija-at, line 52: This bash magic makes my head spin, so I'll just ask: what happens if `npm-publish-order.txt` is missing or empty?

I believe the correct behaviour should be to fallback to the previous logic.

lib/publishing/npm/publish-npm.sh
```diff
@@ -40,8 +40,20 @@ if echo ${TOKENS} | grep 'EAUTHUNKNOWN' > /dev/null; then
...
+
 found=false
-for TGZ in $(find ${PWD}/js -iname '*.tgz'); do
+for filename in $list_of_tarballs; do
```
nija-at, line 55: ooh. what about tarballs in the js/ folder but not in the publish order file? We should publish those, right?

lib/publishing/npm/publish-npm.sh
rix0rrr: I would hope your question would have been pre-empted by the preceding comment. 

Can you tell me what about the comment was not clear?

lib/publishing/npm/publish-npm.sh
rix0rrr: I would hope your question would have been pre-empted by the preceding comment. 

Can you tell me what about the comment was not clear?

lib/publishing/npm/publish-npm.sh
nija-at: My bad.

nija-at (approved):

PR #810, opened 2021-05-11
feat(pipeline): improved pipeline action failure metrics

njlynch:
This extends the new pipeline failure metric approach introduced in #696,
by adding convenience methods to expose pipeline failure metrics at the per-
action level.

-----


test/pipeline.test.ts
```diff
@@ -324,6 +273,115 @@ test('CodeBuild Project name is left undefined when neither buildProjectName nor
   }));
 });
 
+test('metricFailures', () => {
```
nija-at, line 276: Thanks for adding this test too.

nija-at (approved):

PR #1033, opened 2021-09-06
chore: switch delivlib to use projen for publishing

njlynch:
Currently, delivlib publishes via a delivlib pipeline (so meta!). As part of
credential standardization, switching to a projen-based publishing model with
GitHub-managed secrets.

-----


iliapolo:
@njlynch I'm not opposed to tearing down the global delivlib instance - but what would be the story around testing delivlib now? I'd still like to preserve the ability to deploy a delivlib pipeline on my personal account and test it out - are we loosing this/making it much harder?

eladb (approved):

njlynch:
> I'm not opposed to tearing down the global delivlib instance - but what would be the story around testing delivlib now? I'd still like to preserve the ability to deploy a delivlib pipeline on my personal account and test it out - are we loosing this/making it much harder?

The testing story is still the same, as documenting in the contributing guide: https://github.com/cdklabs/aws-delivlib/blob/8b6eac1947b73ca546810e8a66bb493be98f60bd/CONTRIBUTING.md#testing. The instructions are for deploying a delivlib pipeline specifically for test and verification.

PR #1035, opened 2021-09-06
chore: fix package.sh for publishing

njlynch:
The `release` and `bump` tasks write to `dist/changelog.md` and
`dist/version.txt`. Unfortunately, the `publish.sh` script that delivlib
uses removes the `dist` directory before copying over the artifacts. This caused
the release to fail.

Also, explicitly enable publishing to npm

-----

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-set -euo pipefail
-tarball=$(npm pack)
-mkdir -p dist/js
-mv ${tarball} dist/js/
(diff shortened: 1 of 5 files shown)


package.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
```
eladb, line 1: I think this script can be deleted but you'll have to delete lines 71 and 72 in projenrc

eladb (changes requested):

package.sh
njlynch: Done.

iliapolo (approved):

PR #1045, opened 2021-10-14
fix(ecr-mirror): allow using the same directory twice as MirrorSource

rix0rrr:
As long as the build arguments are different.

-----

### DIFF
--- a/lib/registry-sync/mirror-source.ts
+++ b/lib/registry-sync/mirror-source.ts
@@ -115,7 +115,7 @@ export abstract class MirrorSource {
       }
 
       public bind(options: MirrorSourceBindOptions) {
-        const asset = new s3Assets.Asset(options.scope, `BuildContext${this.directory}`, { path: this.directory! });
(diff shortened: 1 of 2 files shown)


lib/registry-sync/mirror-source.ts
```diff
@@ -115,7 +115,7 @@ export abstract class MirrorSource {
       }
 
       public bind(options: MirrorSourceBindOptions) {
-        const asset = new s3Assets.Asset(options.scope, `BuildContext${this.directory}`, { path: this.directory! });
+        const asset = new s3Assets.Asset(options.scope, `BuildContext${this.directory}${JSON.stringify(opts.buildArgs ?? {})}`, { path: this.directory! });
```
nija-at, line 118: Minor: How about this to not change existing construct ids?

```suggestion
        const args = opts.buildArgs ? JSON.stringify(opts.buildArgs) : '';
        const asset = new s3Assets.Asset(options.scope, `BuildContext${this.directory}${args}`, { path: this.directory! });
```

PR #1098, opened 2022-03-28
fix(csr): impossible to produce CSR with keysize > 2KB

RomainMuller:
The Custom Resource used to return the CSR and self-signed certificates
as part of the output payload, but this easily exceeds the 4KB limit
imposed by CloudFormation.

In order to address that, we are now outputting to a specific S3 bucket
instead, and returning pointers to that.

-----


Chriscbr (approved):
Pair programming ftw :)

PR #1101, opened 2022-04-10
chore(integ): allow running integ tests on personal accounts

iliapolo:
Currently, the integration tests can only run in our internal shared account. This isn't good practice, as it prevents multiple developers working on the library at the same time. 

Modify the test so that it can be executed (given some prerequisites) on any account.   

-----

### DIFF
--- a/lib/__tests__/test-stack.ts
+++ b/lib/__tests__/test-stack.ts
@@ -4,7 +4,10 @@ import {
   aws_events as events,
   aws_iam as iam,
   aws_kms as kms,
+  aws_secretsmanager as secrets,
+  aws_ssm as ssm,
 } from 'monocdk';
+import { LinuxBuildImage } from 'monocdk/lib/aws-codebuild';
 import * as delivlib from '../../lib';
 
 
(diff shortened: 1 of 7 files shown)


lib/__tests__/test-stack.ts
```diff
@@ -30,9 +71,12 @@ export class TestStack extends Stack {
     // BUILD
     //
 
+    const superchain = LinuxBuildImage.fromDockerRegistry('public.ecr.aws/c4i6u9i2/jsii/superchain:1-buster-slim-node14');
```
iliapolo, line 74: Unrelated, but changed this to avoid DockerHub throttling. Though I just now realized the image hasn't been updated in about a month...

@RomainMuller Have we stopped publishing to ECR?

See https://gallery.ecr.aws/c4i6u9i2/jsii/superchain

lib/publishing/nuget/publish.sh
```diff
@@ -72,7 +72,7 @@ log=$(mktemp -d)/log.txt
 found=false
 for NUGET_PACKAGE_PATH in $(find dotnet -name *.nupkg -not -iname *.symbols.nupkg); do
     found=true
-    if [ -n "${CODE_SIGNING_SECRET_ID:-}" ]; then
+    if [[ -n "${CODE_SIGNING_SECRET_ID:-}" && "${FOR_REAL:-}" == "true" ]]; then
```
iliapolo, line 75: The self-signed certificate that is dynamically generated in the test is not usable for signing. Figured its ok for dry runs to skip this step.

lib/shellable.ts
```diff
@@ -268,8 +268,6 @@ export class Shellable extends Construct {
       build: this.platform.buildCommands(props.entrypoint),
     }).merge(props.buildSpec || BuildSpec.empty());
 
-    const environmentSecretsAsSecretNames = this.convertEnvironmentSecretArnsToSecretNames(props.environmentSecrets);
```
iliapolo, line 271: Ever since https://github.com/aws/aws-cdk/pull/13706, we actually don't need this since our construct can accept an ARN as well. This was actually causing problems since there was conflicting logic, preventing a secret arn from being a token.

Namely, this was converting the values from an ARN to the secret name, which also ended up in the policy statement, but policy statements must be provided with an ARN.

lib/__tests__/test-stack.ts
```diff
@@ -14,14 +17,52 @@ export class TestStack extends Stack {
...
+    const pypiSecret = new secrets.Secret(this, 'PypISecret', {
+      generateSecretString: {
+        generateStringKey: 'password',
+        secretStringTemplate: JSON.stringify({ username: '__token__' }),
+      },
+    });
+
+    // these need to be real since they are used before the actual publishing action.
+    const tokenSecret = new secrets.Secret(this, 'TokenSecret1', {
+      secretStringBeta1: secrets.SecretStringValueBeta1.fromToken(ssm.StringParameter.valueForStringParameter(this, 'github-token')),
```
iliapolo, line 52: I tried using a secure SSM parameter, but its still not supported for `SecretsManager`.

See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#template-parameters-dynamic-patterns-resources

Would like some input here if you consider this a blocker.

PR #1103, opened 2022-04-15
feat: custom secret keys

iliapolo:
Currently our publishing scripts have either hard coded secret keys, or don't support keys at all and assume the value is the entire secret string. This blocks us from being able to consolidate values in a single secret.

This PR adds keys support for all our publishing scripts. 

Still WIP since it hasn't been thoroughly tested. 

-----

### DIFF
--- a/lib/repo.ts
+++ b/lib/repo.ts
@@ -191,14 +191,14 @@ export interface WritableGitHubRepoProps extends GitHubRepoProps {
    * This is required if you wish to be able to use actions that write to the repo
    * such as docs publishing and automatic bumps.
    */
-  tokenSecret: ExternalSecret;
+  sshKeySecret: ExternalSecret;
 
   /**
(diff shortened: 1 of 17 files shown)


lib/repo.ts
```diff
@@ -189,6 +193,13 @@ export interface WritableGitHubRepoProps extends GitHubRepoProps {
    */
   sshKeySecret: ExternalSecret;
 
+  /**
+   * The key inside the secret holding the ssh key.
+   *
+   * @default - no key. ssh key is assumed to be the entire secret string.
+   */
+  sshKeySecretKey?: string;
```
rix0rrr, line 201: Can we add something like `jsonKey` to `ExternalSecret` instead?

rix0rrr (changes requested):

kaizencc:
Lol github actions bot being extra aggressive these days...

Issue #1083, opened 2022-02-05
Cannot use import statement outside a module


philcali:
I've been trying to use these constructs in my CDK infra, but repeatedly run into the following trace on all versions of `13.x.x`:

```
/node_modules/aws-delivlib/lib/auto-build.ts:1
import {
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at compileFunction (<anonymous>)
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .ts] (/home/philcali/.npm/_npx/1bf7c3c15bf47d04/node_modules/ts-node/src/index.ts:1361:43)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
Subprocess exited with error 1
```

### Environment

- node: 16.13.1
- npm: 8.3.2
- uname: `Linux localhost 5.4.157-17191-g6575d22de694 #1 SMP PREEMPT Thu Jan 20 19:50:13 PST 2022 x86_64 x86_64 x86_64 GNU/Linux` 

### Steps to reproduce

- In a clean workspace using the latest `aws-cdk` tool, init a new app:
```
cdk init app --language=typescript
```
- Update the package.json to match transitive deps for this package:
```
  "devDependencies": {
    "@types/jest": "^27.4.0",
    "@types/node": "^12",
    "aws-cdk": "^1.143.0",
    "jest": "^27.4.7",
    "ts-jest": "^27.1.3",
    "typescript": "~4.5.5"
  },
  "dependencies": {
    "aws-delivlib": "^13",
    "constructs": "^3.3.208",
    "monocdk": "^1.143.0",
    "source-map-support": "^0.5.16"
  }
```
- Update `bin/*.ts` and `lib/*-stack.ts` to use `monocdk` instead of `aws-cdk-lib`.
- Add `delivlib` pipeline to the stack:
```
    new delivlib.Pipeline(this, 'ConsutructsPipeline', {
        repo: new delivlib.GitHubRepo({
            repository: 'username/yourRepo',
[...]

PR #1140, opened 2022-06-10
chore: stop shipping original sources

RomainMuller:
The .npmignore inadvertently included the original .ts files
when those should have been excluded. They cause issues such as
the one described in #1083.

Also excluded other assets that should not have shipped, such
as cdk.out and cdk.json, build scripts.

-----


.projenrc.js
```diff
@@ -108,4 +108,23 @@ const bundlePackageIntegrity = project.addTask('bundle:package-integrity', {
 
 project.compileTask.spawn(bundlePackageIntegrity);
 
+// The npmignore file includes original source files, which is undesirable.
+project.npmignore.exclude(
+  '/lib/**/*.ts',
+);
+project.npmignore.include(
```
Naumel, line 115: Should this be `exclude` to match the pattern above?

Naumel (approved):

PR #1145, opened 2022-07-14
fix: Access githubtoken w/ unsafeUnwrap in AutoBuild

biffgaut:
Directly accessing the secret githubtoken in AutoBuild is no longer allowed. Code must explicitly call unsafeUnwrap() to confirm the developer understands the data value is a secret and the risks associated with accessing it.

-----


iliapolo:
@biffgaut just to make sure - you're application has the `@aws-cdk/core:checkSecretUsage` feature flag turned on right?

biffgaut:
Yes

```
    "@aws-cdk/core:checkSecretUsage": true,

```

iliapolo (approved):

PR #1261, opened 2023-03-24
fix: NuGetPublish reads the certificate incorrectly

otaviomacedo:
The value of the SSM parameter is the URL of the S3 bucket that contains the certificate (`s3://delivlib-test-x509codesigningkeyrsaprivatekeycert-11xb1jecarfb/self-signed-certificate.pem`), rather than the certificate itself.

Add a step to download the certificate from the bucket, and set up the corresponding policies.

-----

### DIFF
--- a/lib/publishing.ts
+++ b/lib/publishing.ts
@@ -245,17 +245,24 @@ export class PublishToNuGetProject extends Construct implements IPublisher {
 
     environment.NUGET_SECRET_ID = props.nugetApiKeySecret.secretArn;
 
-    if (props.codeSign) {
-      environment.CODE_SIGNING_SECRET_ID = props.codeSign.credential.secretArn;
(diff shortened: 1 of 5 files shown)


lib/publishing.ts
```diff
@@ -245,18 +245,25 @@ export class PublishToNuGetProject extends Construct implements IPublisher {
...
       scriptDirectory: path.join(__dirname, 'publishing', 'nuget'),
       entrypoint: 'publish.sh',
       environment,
     });
 
+    if (props.codeSign) {
+      environment.CODE_SIGNING_SECRET_ID = props.codeSign.credential.secretArn;
+      environment.CODE_SIGNING_PARAMETER_NAME = props.codeSign.principal.parameterName;
+
+      if (props.codeSign.certificateBucket) {
```
rix0rrr, line 259: This could use `grantRead`

rix0rrr (approved):

PR #1265, opened 2023-03-27
chore: authenticate ECRMirror to ECR Public

RomainMuller:
This affords an increased rate limit for pulls. Additionally, switch to using the `public.ecr.aws/jsii/superchain` image instead of pulling it from DockerHub, and clarify the MirrorImage doesn't need to come from DockerBug by introducing an alternate (and improved) static factory.

In orde to enable reducing the scope of changes incurred by migrating from DockerHub images to using their equivalent ECR-Public URL, allows configuring a custom ECR Repository name to be used on the output side.

-----

### DIFF
--- a/lib/registry-sync/mirror-source.ts
+++ b/lib/registry-sync/mirror-source.ts
@@ -65,8 +65,22 @@ export abstract class MirrorSource {
    *
    * @param image e.g jsii/superchain
    * @param tag optional, defaults to 'latest'
+   *
+   * @deprecated This method's name inaccurately expresses that the image comes
+   * from DockerHub, when any publicly-accessible repository can be used. Prefer
+   * using `fromImageName(string, string?)` instead, which is more aptly named.
    */
   public static fromDockerHub(image: string, tag: string = 'latest'): MirrorSource {
+    return this.fromImageName(image, tag);
+  }
+
+  /**
+   * Configure an image from DockerHub or a repository-qualified image name.
+   *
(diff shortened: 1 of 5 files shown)


lib/registry-sync/mirror-source.ts
```diff
@@ -65,26 +65,41 @@ export abstract class MirrorSource {
...
+  }
+
+  /**
+   * Configure an image from DockerHub or a repository-qualified image name.
+   *
+   * @param image e.g public.ecr.aws/jsii/superchain
+   * @param tag optional, defaults to 'latest'
+   * @param ecrRepositoryName the name of the ECR Repository to use (e.g: jsii/superchain)
+   */
+  public static fromImageName(image: string, tag: string = 'latest', ecrRepositoryName: string = image.includes('/') ? image : `library/${image}`): MirrorSource {
```
iliapolo, line 84:
```suggestion
  public static fromImageName(image: string, tag: string = 'latest', ecrRepositoryName?: string): MirrorSource {
```

Is this really needed? I see the constructor already defaults `ecrRepositoryName` to `repositoryName`, which seems like that results in the same thing?

Naumel (approved):

lib/registry-sync/mirror-source.ts
RomainMuller: Yep - the default `library/` prefix is only actioned on that code path, if no custom name is provided... The other cases just use the exact same image name that is provided by default... This was warranted to not change the behavior of existing unit tests.

PR #1466, opened 2023-10-04
feat: sign stage for NuGet code signing with AWS Signer

colifran:
Currently, we (the AWS CDK team) sign `.nupkg` files using `lib/publishing/nuget/sign.sh` with certificates we maintain in our cdk-ops repo. We received guidance from another internal AWS team that this no longer conforms with AWS best practice.

The new mechanism for code signing `.nupkg` files will be carried out using AWS Signer with infrastructure set up in a signing account. This PR adds a signing stage feature and a new `sign.sh` script that will be used to interact with our signing account to code sign `.nupkg` files.

-----


lib/pipeline.ts
```diff
@@ -357,6 +359,21 @@ export class Pipeline extends Construct {
...
+    signer.addToPipeline(stage, `${signer.node.id}Sign`, {
+      inputArtifact: options.inputArtifact || this.buildOutput,
+      runOrder: this.determineRunOrderForNewAction(stage),
+    });
+  }
+
+  public signNuGetWithSigner(options: signing.SignNuGetWithSignerProps & signing.AddSigningOptions) {
+    this.addSigning(new signing.SignNuGetWithSigner(this, 'NuGetSigning', {
+      ...options,
+    })), options;
```
rix0rrr, line 374:
```suggestion
    }));
```

lib/signing.ts
```diff
@@ -0,0 +1,90 @@
...
+      entrypoint: 'sign.sh',
+      environment,
+    });
+
+    this.role = shellable.role;
+    this.project = shellable.project;
+  }
+
+  public addToPipeline(stage: IStage, id: string, options: AddToPipelineOptions) {
+    stage.addAction(new CodeBuildAction({
```
rix0rrr, line 83: I think we probably also need to have an output artifact. Otherwise, how are we going to get to the signed artifact?

rix0rrr: And we'll need to have an artifact configuration in the shellable (means you need to supply a [buildspec](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#build-spec-ref-example), which most examples will say needs to go into the input artifact/source repo, but in our case has to go onto the `Project` directly).

rix0rrr (changes requested):

lib/pipeline.ts
colifran: Whoops! Fixed.

lib/signing.ts
colifran: Updated the code with the ability to get the output artifact from the signing step. I also added code to to add an artifact configuration to to the buildspec of the signing stage. I mirrored what I saw in the buildspec for the build stage since the signing stage is accepting the output artifact from the build stage and signing the dll files in the nupkg. Let me know your thoughts.

lib/signing.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const shellable = new Shellable(this, 'Default', {
+      platform: new LinuxPlatform(props.buildImage ?? LinuxBuildImage.fromDockerRegistry('public.ecr.aws/jsii/superchain:1-buster-slim-node18')),
+      scriptDirectory: path.join(__dirname, 'signing', 'nuget'),
+      entrypoint: 'sign.sh',
+      buildSpec: BuildSpec.literal({
+        version: '0.2',
+        artifacts: {
+          files: ['**/*'],
+          ['base-directory']: '../src',
```
rix0rrr, line 80: Why `../src`? Does that not bring us back to right here? (i.e. `.`?)

rix0rrr (approved):

lib/signing.ts
colifran: Yeah. For some reason during my testing I thought using `.` was expanding all the sub-directories and not maintaining the file structure, i.e., `dotnet`, `python`, `java`, `go`, etc. I just re-tested using `.` as the `base-directory` and it works the same way as using `../src`.

PR #1476, opened 2023-10-12
feat: expose service role to allow custom codebuild role for shellables

colifran:
This PR exposes the underlying `role` property used to specify a role to be used by the CodeBuild project created as part of creating a new `Shellable`

-----

### DIFF
--- a/lib/shellable.ts
+++ b/lib/shellable.ts
@@ -6,6 +6,7 @@ import {
   aws_codepipeline as cpipeline, aws_codepipeline_actions as cpipeline_actions,
   aws_iam as iam, aws_s3_assets as assets, aws_secretsmanager, aws_ssm,
 } from 'aws-cdk-lib';
+import { IRole } from 'aws-cdk-lib/aws-iam';
 import { Construct } from 'constructs';
 import { BuildSpec } from './build-spec';
 import { renderEnvironmentVariables } from './util';
@@ -153,6 +154,13 @@ export interface ShellableOptions {
    * @default No environment variables
    */
   readonly pipelineEnvironmentVars?: Record<string, string>;
+
+  /**
+   * The service role to assume while running the build
+   *
+   * @default A role will be created
+   */
+  readonly serviceRole?: IRole;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


lib/shellable.ts
```diff
@@ -153,6 +154,13 @@ export interface ShellableOptions {
    * @default No environment variables
    */
   readonly pipelineEnvironmentVars?: Record<string, string>;
+
+  /**
+   * The service role to assume while running the build
+   *
+   * @default A role will be created
```
kaizencc, line 161: where is "a role will be created" happening?

colifran: This happens in the codebuild `Project` class - it's the behavior that happens currently. See: https://github.com/aws/aws-cdk/blob/8557c40ce3efb16cb6c6f4677540d38853937166/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1045

kaizencc: Oh right. Okay, so the game plan is to send in our own serviceRole with similar permissions.

kaizencc (approved):

lib/shellable.ts
colifran: Exactly. It would look something like this:
```ts
private addSigning() {
    const signingBucket = Bucket.fromBucketArn(this, 'SigningBucket', config.signingBucketArn);
    const signingLambda = Function.fromFunctionArn(this, 'SigningFunction', config.signingLambdaArn);
    const accessRole = Role.fromRoleArn(this, 'AccessRole', config.signingAccessRoleArn);
    const serviceRole = new Role(this, 'SeviceRole', {
      roleName: 'signing-codebuild-role',
      assumedBy: new ServicePrincipal('codebuild.amazonaws.com'),
    });
    serviceRole.addToPolicy(new PolicyStatement({
      actions: ['sts:AssumeRole'],
      resources: [accessRole.roleArn],
    }));

    this.pipeline.signNuGetWithSigner({
      signingBucket,
      signingLambda,
      accessRole,
      serviceRole,
    });
  }
```

PR #1477, opened 2023-10-12
feat: add props on `SignNuGetWithSignerProps` for signer profile and signer owner

colifran:
This PR adds properties on the `SignNuGetWithSignerProps` interface that allow a user to specify the profile and owner that will be used for signing. These properties are added as environment variables that are passed in the payload to the signing lambda as part of the`sign.sh` script. The signing lambda can utilize these values to start the signing job with the specified signer profile and owner. In addition to the above changes, `sign.sh` has been updated to provided the entire path of the dll file to upload to the signing bucket. This fixes an error where the --body value must be a full path.

-----

### DIFF
--- a/lib/signing.ts
+++ b/lib/signing.ts
@@ -92,11 +92,11 @@ export class SignNuGetWithSigner extends Construct implements ISigner {
     };
 
     if (props.signerProfileName) {
-      environment.SIGNER_PROFILE_NAME = props.signerProfileName;
+      environment['SIGNER_PROFILE_NAME'] = props.signerProfileName;
     }
 
     if (props.signerProfileOwner) {
-      environment.SIGNER_PROFILE_OWNER = props.signerProfileOwner;
+      environment['SIGNER_PROFILE_OWNER'] = props.signerProfileOwner;
     }
 
     const shellable = new Shellable(this, 'Default', {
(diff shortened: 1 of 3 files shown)


lib/signing.ts
```diff
@@ -67,6 +81,8 @@ export class SignNuGetWithSigner extends Construct implements ISigner {
       SIGNING_BUCKET_NAME: props.signingBucket.bucketName,
       SIGNING_LAMBDA_ARN: props.signingLambda.functionArn,
       ACCESS_ROLE_ARN: props.accessRole.roleArn,
+      SIGNER_PROFILE_NAME: props.signerProfileName ?? 'profile-name',
+      SIGNER_PROFILE_OWNER: props.signerProfileOwner ?? 'profile-owner',
```
kaizencc, line 85: i think if not specified these env variables should be undefined. Any reason this is not the case?

colifran: I don't have a specific reason besides thinking maybe we should try for starting the signing job with profile-name and profile-owner if nothing is explicitly provided. I can update this to make it undefined if nothing is provided.

colifran: This has been updated.

kaizencc (approved):

PR #1478, opened 2023-10-12
fix: `sign.sh` incorrectly uses relative path of nuget package instead of absolute path for signing operations

colifran:
`sign.sh` was incorrectly using the relative nuget package path rather than the absolute nuget package path for signing operations. This was causing the signing process to exit with status 12. This PR corrects this mistake by getting the absolute nuget package path via the following code:

```
nuget_package=$(cd $(dirname ${nuget_package_path}) && echo $PWD)/$(basename ${nuget_package_path})
```

-----


lib/signing/nuget/sign.sh
```diff
@@ -21,12 +21,13 @@ fi
 found=false
 for nuget_package_path in $(find dotnet -name *.nupkg -not -iname *.symbols.nupkg); do
   found=true
-  echo "🔑 Applying authenticode signatures to assemblies in ${nuget_package_path}"
-  for file in $(unzip -Z1 ${nuget_package_path} '*.dll'); do
+  nuget_package=$(cd $(dirname ${nuget_package_path}) && echo $PWD)/$(basename ${nuget_package_path})
```
rix0rrr, line 24: It looks like what you're trying to do is make this path absolute.

Most Linuxes will have the tool `realpath` to do this (won't exist on Mac by default though), and according to StackOverflow you can use `readlink -f <file>` as well.

rix0rrr (approved):
Approved but maybe document what and why you're doing a bit more precisely.

And update the title: you need the absolute path because you are `cd`ing in a subshell, so a relative path name would change meaning when you do that.

lib/signing/nuget/sign.sh
colifran: Noted - thanks for explaining that and showing me an alternative!

PR #1550, opened 2024-01-25
chore: upgrade minNodeVersion to 18+

kaizencc:
-----

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16.17.0
+          node-version: 18.x
       - name: Install dependencies
         run: yarn install --check-files
       - name: build
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -9,6 +9,7 @@ const project = new typescript.TypeScriptProject({
   authorName: 'Amazon Web Services',
   authorUrl: 'https://aws.amazon.com',
   minNodeVersion: '18.12.0',
+  workflowNodeVersion: '18.x',
   keywords: [
     'aws-cdk',
     'continuous-delivery',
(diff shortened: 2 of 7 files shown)


.projenrc.ts
```diff
@@ -8,7 +8,7 @@ const project = new typescript.TypeScriptProject({
   defaultReleaseBranch: 'main',
   authorName: 'Amazon Web Services',
   authorUrl: 'https://aws.amazon.com',
-  minNodeVersion: '16.17.0',
+  minNodeVersion: '18.12.0',
```
kaizencc, line 11:
```suggestion
  minNodeVersion: '18.12.0',
  workflowNodeVersion: '18.x',
```

.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
```
jfuss, line 23: Looks like we are duplicating work: https://github.com/cdklabs/aws-delivlib/pull/1551

You should be upgrading setup-node and checkout versions as well.

jfuss (changes requested):

.github/workflows/build.yml
kaizencc: Sorry. Do you want to take over? On your PR, it looks like you manually updated the files whereas you'll need to update the `projenrc.ts` file for any changes to stick.

kaizencc: These versions are controlled by projen. I don't think they are necessary to fix the broken workflow atm

jfuss: There isn't much to take over here...

A little confused why projen is needed here but it's how it is setup.

The error (which I can't find anymore) said checkout and node actions needed upgrading to node v20. Which I don't think CDK even supports.

jfuss (approved):

PR #1589, opened 2024-03-05
chore: remove preinstall from windows

TheRealAmazonKendra:
All supported CodeBuild versions of Windows now come with node preinstalled. The download process frequently fails. Removing it will reduce noise.

-----


mrgrain:
Wondering if this is more than a chore though. Almost feels like a breaking change?! But then I'm not sure what the relationship between delivib and the used images is.

PR #1731, opened 2024-09-13
feat: remove the lambda layers used for custom resources underneath lambda functions

moelasmar:
This change removed the Lambda layers we used to add the pgp and openssl binaries to the lambda functions that are used to implement the OpenPGPKeyPair, RsaPrivateKeySecret, CodeSigningCertificate, and CertificateSigningRequest constructs.

The change is to make these lambda functions to be of Image package type, so we can install the required binaries in the docker image, instead of maintaining the binaries in the github repo.

I followed this section https://github.com/cdklabs/aws-delivlib/blob/main/CONTRIBUTING.md#testing for testing these changes, and mainly the part of running `yarn integ:update`

-----


lib/custom-resource-handlers/pgpSecretDockerfile
```diff
@@ -0,0 +1,16 @@
+# Use a NodeJS 20.x runtime
+FROM public.ecr.aws/lambda/nodejs:20-x86_64
+
+# install openssel and
+RUN dnf install -y openssl
+ENV LD_LIBRARY_PATH=""
```
shikha372, line 6: to understand, why we have this as an empty string ?

lib/custom-resource-handlers/src/certificate-signing-request.ts
```diff
@@ -60,12 +60,12 @@ async function _createSelfSignedCertificate(event: cfn.Event): Promise<ResourceA
     const configFile = await _makeCsrConfig(event, tempDir);
     const pkeyFile = await _retrievePrivateKey(event, tempDir);
     const csrFile = path.join(tempDir, 'csr.pem');
-    await _exec('/opt/openssl', 'req', '-config', configFile,
+    await _exec('openssl', 'req', '-config', configFile,
```
shikha372, line 63: is it due to any version change, that the path is modified now or it could've been this way earlier as well

moelasmar: I found that openssl is installed in /bin, but in general we do not need to determine the full path

lib/custom-resource-handlers/pgpSecretDockerfile
moelasmar: I faced the issue mentioned in [this thread](https://stackoverflow.com/questions/77527321/openssl-command-not-found-on-amazon-linux-2023), and so I applied the suggested solution, and it worked for me

lib/custom-resource-handlers/certificateSigningRequestDockerfile
```diff
@@ -0,0 +1,16 @@
+# Use a NodeJS 20.x runtime
+FROM public.ecr.aws/lambda/nodejs:20-x86_64
+
+# install openssel and
+RUN dnf install -y openssl
+ENV LD_LIBRARY_PATH=""
+RUN openssl version
+
+# install gpg-agent
+RUN rpm -e gnupg2-minimal --nodeps && dnf -y install gnupg2-full && gpg-agent --daemon
```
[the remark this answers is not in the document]
comcalvi: `-e` removes the package, so `-e --no-deps` means "remove only the package without removing its dependencies"

lib/custom-resource-handlers/privateKeyDockerfile
```diff
@@ -0,0 +1,15 @@
+FROM public.ecr.aws/lambda/nodejs:20-x86_64
+
+# install openssel and
```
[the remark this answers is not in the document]
comcalvi: Discussed offline; can use an env var to change the one difference between them

lib/__tests__/test-stack.ts
```diff
@@ -142,7 +142,8 @@ export class TestStack extends Stack {
       identity: 'aws-cdk-dev',
       keySizeBits: 4_096,
       pubKeyParameterName: `/${this.node.path}/CodeSign.pub`,
-      secretName: this.node.path + '/CodeSign',
+      // I am not able to immediately delete the existing secret, it needs to wait at least 7 days.
```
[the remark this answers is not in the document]
comcalvi: The whole chain of resource replacements is:

1. We need to change the custom resource handler's packaging type from zip to Docker image, but Lambda doesn't allow that; we need to make a new Function, which we do by changing the logical ID.
2. Changing the Custom Resource Service Token is similarly forbidden, so we must also create a new one; this is another logical ID change. 
3. Recreating the custom resource will cause it to create a new key, which will fail because the key with that name (`this.node.path + '/CodeSign'`) already exists; so, we must change the key name.

@moelasmar I would think that step 3 would cause the secret to be deleted as well, because the custom resource should delete it on the `DELETE` event. Why doesn't it?

lib/__tests__/test-stack.ts
```diff
@@ -185,13 +184,6 @@ export class TestStack extends Stack {
       bumpCommand: 'npm i && npm run bump',
     });
 
-    //
-    // AUTO-BUILD
-
-    pipeline.autoBuild({
-      publicLogs: true,
-    });
```
comcalvi, line 192: Why the deletion here? We want this to be tested, no?

comcalvi (approved):

PR #1773, opened 2024-11-18
feat: write publishing timestamps to SSM

rix0rrr:
Add a feature to allow publishing the latest published version and its timestamp to SSM, so that we can start tracking how long it takes to replicate, and alarm at a different severity if a canary fails and we recently released.

-----


lib/publishing/github/update-ssm.sh
otaviomacedo, line 1: Can't we have this script in a single place?

rix0rrr: Not trivially -- the way the assets are packed means it just uses the specific publishing directory.

We could go for a build step to copy a bunch of files into a specific directory, but since bash doesn't have a way to find or specify directories, it would all be quite magical. The publishing scripts would invoke another script that's not there from the PoV of the calling script (but will be there at runtime because it will have been copied)

otaviomacedo: Ah, I see the problem. Would it make sense to at least have a build step to make sure they are all identical?

rix0rrr: Yeah I can do that.

otaviomacedo (approved):

PR #1799, opened 2024-12-09
fix: bump default superchain images

rix0rrr:
delivlib has some places where it defaults to a particular superchain image, which may grow outdated over time. This came up specifically in the context of ECR mirroring recently.

Bump the Superchain image we're using to a recent one everywhere (`bookworm` instead of `bullseye`, a more recent Node because Node 18 is about to be deprecated).

Also make sure that all CodeBuild projects have reasonable descriptions, because we have hundreds in our account now and we need some way to tell them apart.

-----

### DIFF
--- a/lib/__tests__/registry-sync/ecr-mirror.test.ts
+++ b/lib/__tests__/registry-sync/ecr-mirror.test.ts
@@ -7,6 +7,7 @@ import {
 } from 'aws-cdk-lib';
 import { Template, Match } from 'aws-cdk-lib/assertions';
 import { EcrMirror, EcrMirrorAspect, MirrorSource } from '../../../lib/registry-sync';
+import { DEFAULT_SUPERCHAIN_IMAGE } from '../../constants';
 
 describe('EcrMirror', () => {
   test('default', () => {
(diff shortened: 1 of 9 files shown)


lib/__tests__/registry-sync/ecr-mirror.test.ts
```diff
@@ -36,7 +37,7 @@ describe('EcrMirror', () => {
             Value: '123aass:password-key:AWSCURRENT',
           },
         ],
-        Image: 'public.ecr.aws/jsii/superchain:1-bullseye-slim-node18',
+        Image: DEFAULT_SUPERCHAIN_IMAGE,
```
otaviomacedo, line 40: Are we really testing anything here, now? If we set `DEFAULT_SUPERCHAIN_IMAGE` to `"some-nonsensical-string"`, it will still pass.

rix0rrr: True. Copying the string there feels just as useless, but I can do it I suppose

iliapolo (approved):

PR #2062, opened 2026-09-08
feat: support assumeRole on Windows

dgandhi62:
Replicates the Linux `assumeRole` prebuild logic to PowerShell so that `WindowsPlatform`-backed `Shellable` actions can assume an IAM role. Previously any truthy `assumeRole` on Windows threw
at synth time: assumeRole is not supported on Windows: https://github.com/cdklabs/aws-delivlib/issues/57

That throw was introduced in #47 (Jan 2019), which shipped assume-role for Linux only and deferred Windows as tracked work. Issue #57 was later auto-closed by the stale bot without the feature ever being implemented, so no published version supports Windows assume-role today. This PR implements it.

-----


mrgrain (approved):

lib/__tests__/test-stack.ts
```diff
@@ -84,6 +84,26 @@ export class TestStack extends Stack {
       },
     });
 
+    // Windows equivalent of the AssumeRole test above. Gated behind the same
+    // WINDOWS flag as HelloWindows because Windows CodeBuild is slow (~10 min).
+    // Flip WINDOWS to true and run `yarn integ:update` to verify that assumeRole
```
iankhou, line 89: Is flipping WINDOWS to `true` meant as a manual step during development? I see we changed it to `true` and committed it. Either the comment here is wrong, or the set value of WINDOWS is incorrect.

lib/shellable.ts
```diff
@@ -576,17 +576,53 @@ export class WindowsPlatform extends ShellPlatform {
     ];
   }
 
-  public prebuildCommands(assumeRole?: AssumeRole, _useRegionalStsEndpoints?: boolean): string[] {
+  public prebuildCommands(assumeRole?: AssumeRole, useRegionalStsEndpoints?: boolean): string[] {
+    const lines = new Array<string>();
+
```
iankhou, line 581: On Linux, we download scripts from S3. On Windows, this step looks like it's done during the build step. Why the distinction? They end up using different credentials.

lib/__tests__/test-stack.ts
dgandhi62: Ubuntu tests are not gated at all, they always run. For windows, if WINDOWS = true, the pipeline created adds the windows tests as a codebuild job. if its False, it always ignores windows tests. 

The reason it was false before (in my opinion), is simply because we didn't support windows and therefore the fact that it took more time during development was relevent. We do now. Ideally, we want to be removing this gating altogether but I kept it in for the time being

lib/shellable.ts
dgandhi62: Ah good catch. I didn't notice this...we already had existing code in build phase for the windows s3 download. Because of this change, we assume roles first during pre-build and the download runs later in build (which makes no sense). I'll address this.

